PR #26, opened 2020-07-08
Modify README file for charts

### LINKED ISSUE #24: Update tl;dr instructions for installing the Chart from the 'target'
The Chart includes a README with the instructions of how to get & install the Chart. 

e.g.

```
$ helm repo add bitnami https://charts.bitnami.com/bitnami
$ helm install my-release bitnami/apache
```

Ideally we should replace those instructions to use the "target" Chart repository (instead of the "source" one) so users can see those instructions.

tompizmor:
Fix #24 

Bitnami charts contain the following section in the README:

```
## TL;DR;
$ helm repo add bitnami https://charts.bitnami.com/bitnami
$ helm install my-release bitnami/drupal
```

As we are migrating the chart to a new registry, it makes sense to update these instructions in the README.

The expected result should be:

```
## TL;DR;
$ helm repo add $REPO_NAME $TARGET_URL
$ helm install my-release $REPO_NAME/drupal
```

Where `$REPO_NAME` is specified in the config file via the property `target.repoName` or 'myrepo' by default if it is not specified, and` $TARGET_URL` is specified via `target.repo.url`


pkg/chart/chart.go
```diff
@@ -58,3 +60,30 @@ func updateContainerImageRegistry(valuesFile string, targetRepo *api.TargetRepo)
...
+	// Update helm repo add with string replacement
+	addBitnamiRepoLine := fmt.Sprintf("helm repo add bitnami %s", sourceURL)
+	addCustomRepoLine := fmt.Sprintf("helm repo add %s %s", repoName, targetURL)
+	newContent := strings.ReplaceAll(string(readme), addBitnamiRepoLine, addCustomRepoLine)
+	// Update bitnami/chart references with regex
+	regexString := fmt.Sprintf(`(?m)(\s)(bitnami/%s)(\s)`, chartName)
+	regex := regexp.MustCompile(regexString)
+	submatch := regex.FindStringSubmatch(string(readme))
+	if len(submatch) > 0 {
+		klog.V(2).Infof("Updating bitnami/ references")
```
jotadrilo, line 80: I would use a bigger verbose level, since this is not useful for a regular user

pkg/chart/chart.go
```diff
@@ -58,3 +60,30 @@ func updateContainerImageRegistry(valuesFile string, targetRepo *api.TargetRepo)
...
+	if err != nil {
+		return errors.Trace(err)
+	}
+	// Update helm repo add with string replacement
+	addBitnamiRepoLine := fmt.Sprintf("helm repo add bitnami %s", sourceURL)
+	addCustomRepoLine := fmt.Sprintf("helm repo add %s %s", repoName, targetURL)
+	newContent := strings.ReplaceAll(string(readme), addBitnamiRepoLine, addCustomRepoLine)
+	// Update bitnami/chart references with regex
+	regexString := fmt.Sprintf(`(?m)(\s)(bitnami/%s)(\s)`, chartName)
+	regex := regexp.MustCompile(regexString)
```
jotadrilo, line 77: declare the regex expression as a top-level package variable so it fails at initialization time, not once the program reaches this point

pkg/chart/chart.go
```diff
@@ -58,3 +60,30 @@ func updateContainerImageRegistry(valuesFile string, targetRepo *api.TargetRepo)
...
+	if len(submatch) > 0 {
+		klog.V(2).Infof("Updating bitnami/ references")
+		replaceText := fmt.Sprintf("%s%s/%s%s", submatch[1], repoName, chartName, submatch[3])
+		newContent = regex.ReplaceAllString(newContent, replaceText)
+	}
+	err = ioutil.WriteFile(readmeFile, []byte(newContent), 0)
+	if err != nil {
+		return errors.Trace(err)
+	}
+	return nil
```
jotadrilo, line 88:
```suggestion
	return errors.Trace(ioutil.WriteFile(readmeFile, []byte(newContent), 0))
```

pkg/config/config.go
```diff
@@ -24,6 +29,10 @@ func Load(config *api.Config) error {
 	if err != nil {
 		return errors.Trace(fmt.Errorf("Error unmarshalling config file: %w", err))
 	}
+	if config.Target.RepoName == "" {
+		klog.V(2).Infof("'target.repoName' property is empty. Using 'myrepo' default value")
```
jotadrilo, line 33: this should be either a warning or a bigger verbosity message

README.md
```diff
@@ -51,6 +51,7 @@ source:
     #   username: "USERNAME"
     #   password: "PASSWORD"
 target:
+  repoName: "myrepo"
```
jotadrilo, line 54: I would document what's this setting for (you can do this later on since other settings are not documented either)

tompizmor: yes, I am working on the documentation in a different PR.

pkg/chart/chart.go
tompizmor: note that for this case the regex depends on the chartName

pkg/config/config.go
tompizmor: I think I prefer a warning

pkg/chart/chart.go
jotadrilo: then please consider using [`Compile`](https://golang.org/pkg/regexp/#Compile) rather than [`MustCompile`](https://golang.org/pkg/regexp/#MustCompile) since the former is thought to be used in this kind of case

PR #29, opened 2020-07-14
not capitalize errors and use t.Fatalf if test preparation step fail

### LINKED ISSUE #23: Fix unexpected tests workflow
I have spotted a potential issue in some tests. Example:

https://github.com/bitnami-labs/chart-repository-syncer/blob/2748f016af197f881284b076c7e505bdf044ecfc/pkg/chart/dependency_test.go#L45-L48

This makes the program to mark the error as a test error instead of an actual error in the program.
In theory, all errors produced by the initialization or preparation of the test should be an error program, not a test error.

For example:

```
	testTmpDir, err := ioutil.TempDir("", "c3tsyncer-tests")
	if err != nil {
		t.Fatalf("Error creating temporary: %s", testTmpDir)
	}
```

Additionally, in order to follow some Go patterns, [errors should not be capitalized](https://github.com/golang/go/wiki/CodeReviewComments#error-strings).

This issue is meant to tackle some technical debt and change `t.Errorf` references by `t.Fatalf`, which stops the current test immediately.

tompizmor:
Fix #23 

- errors are now **not** capitalized as recommended for golang projects. https://github.com/golang/go/wiki/CodeReviewComments#error-strings
- use `t.Fatal(f)` instead of `t.Error(f)` if it fails a test preparation step instead of the test itself

### DIFF
--- a/cmd/sync_chart.go
+++ b/cmd/sync_chart.go
@@ -48,7 +48,7 @@ func newSyncChart() *cobra.Command {
 
 func syncChart() error {
 	if !syncAllVersions && version == "" {
-		return errors.Trace(fmt.Errorf("Please specify a version to sync with --version VERSION or sync all versions with --all-versions"))
+		return errors.Trace(fmt.Errorf("please specify a version to sync with --version VERSION or sync all versions with --all-versions"))
 	}
 
 	// Load config file
@@ -65,7 +65,7 @@ func syncChart() error {
 	// Load index.yaml info into index object
 	sourceIndex, err := utils.LoadIndexFromRepo(source.Repo)
 	if err != nil {
-		return errors.Trace(fmt.Errorf("Error loading index.yaml: %w", err))
+		return errors.Trace(fmt.Errorf("error loading index.yaml: %w", err))
 	}
 
 	// Add target repo to helm CLI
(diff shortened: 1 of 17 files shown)


cmd/sync_chart.go
```diff
@@ -48,7 +48,7 @@ func newSyncChart() *cobra.Command {
 
 func syncChart() error {
 	if !syncAllVersions && version == "" {
-		return errors.Trace(fmt.Errorf("Please specify a version to sync with --version VERSION or sync all versions with --all-versions"))
+		return errors.Trace(fmt.Errorf("please specify a version to sync with --version VERSION or sync all versions with --all-versions"))
 	}
```
jotadrilo, line 52: The right way is adding a `PreRunE` hook to the cmd. Example:

```
cmd := &cobra.Command{
    ...
    PreRunE: func(cmd *cobra.Command, args []string) error {
        if !syncAllVersions && version == "" {
            return errors.Trace(fmt.Errorf(`missing "--version" flag: Please use either "--version VERSION" or "--all-versions"))
        }
        if syncAllVersions && version != "" {
            klog.Warningf(`Ignoring "--version" flag: "--all-versions" is set))
        }
        return nil
    },
```

tompizmor: done!

PR #30, opened 2020-07-31
Fix getting auth properties from env vars

### LINKED ISSUE #31: Environment variables used to set auth doesn't work
I am trying to use the tool with a target repository that requires authentication. Using the yaml everything works as expected:
```yaml
target:
   auth:
     # username is the username used to authenticate against the target chart repo
     # `TARGET_AUTH_USERNAME` env var can be used instead of this entry
     username: "robot$chart-syncer"
     # password is the password used to authenticate against the target chart repo
     # `TARGET_AUTH_PASSWORD` env var can be used instead of this entry
     password: "somePasswrod"
```
but when using the env. variables (both `TARGET_AUTH_*` and `TARGET_REPO_*`) with the same values, the authentication doesn't work:
```
POST chart status Code: 401, Message: {"code":401,"message":"UnAuthorized"}
```

tompizmor:
Since we stop using viper to unmarshall the config file the method `viper.BindEnv` stop working too and we lost the ability to set auth properties from env vars.

This PR enables the functionality again and adds a test to verify it continues working on future releases.

Fix #31

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -2,3 +2,24 @@
 
 // Package api provides APIs for syncing a chart repository
 package api
+
+// SetBasicAuth is used configure the auth credentials for Repo kinds
+func (r *Repo) SetBasicAuth(username string, password string) error {
+	// No auth provided, leave as-is
+	if username == "" && password == "" {
+		return nil
+	}
+	// If username or password are already set, the value from
+	// config file has preference over environment variable
+	if r.Auth == nil {
+		r.Auth = &Auth{Username: username, Password: password}
+	} else {
+		if r.Auth.Username == "" {
+			r.Auth.Username = username
+		}
+		if r.Auth.Password == "" {
+			r.Auth.Password = password
+		}
+	}
+	return nil
+}
--- a/api/api.go
+++ b/api/api.go
@@ -9,15 +9,16 @@ func (r *Repo) SetBasicAuth(username string, password string) error {
 	if username == "" && password == "" {
 		return nil
 	}
-	// If username or password are already set, the value from
-	// config file has preference over environment variable
+	// If username or password are already set, the value will
+	// be override with the environment variable one as it
+	// has preference.
 	if r.Auth == nil {
 		r.Auth = &Auth{Username: username, Password: password}
 	} else {
-		if r.Auth.Username == "" {
+		if username != "" {
 			r.Auth.Username = username
 		}
-		if r.Auth.Password == "" {
+		if password != "" {
 			r.Auth.Password = password
 		}
 	}
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
+	"os"
 
 	"github.com/bitnami-labs/charts-syncer/api"
 	"github.com/bitnami-labs/pbjson"
@@ -20,11 +21,6 @@ const (
 
 // Load unmarshall config file into Config struct.
 func Load(config *api.Config) error {
-	viper.BindEnv("source.auth.username", "SOURCE_AUTH_USERNAME")
-	viper.BindEnv("source.auth.password", "SOURCE_AUTH_PASSWORD")
-	viper.BindEnv("target.auth.username", "TARGET_AUTH_USERNAME")
-	viper.BindEnv("target.auth.password", "TARGET_AUTH_PASSWORD")
-
 	err := yamlToProto(viper.ConfigFileUsed(), config)
 	if err != nil {
 		return errors.Trace(fmt.Errorf("error unmarshalling config file: %w", err))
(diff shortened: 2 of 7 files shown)


carrodher:
Thanks for the quick fix!

There are some mentions to `*_REPO_*` env. variables in the README and also in the cronjob.yaml examples, can you replace those ones by `*_AUTH_*`?

pkg/config/config.go
```diff
@@ -33,7 +29,26 @@ func Load(config *api.Config) error {
...
-
+	if config.Source.Repo.Auth == nil {
+		sourceAuth := &api.Auth{}
+		if su := os.Getenv("SOURCE_AUTH_USERNAME"); su != "" {
+			sourceAuth.Username = su
+		}
+		if sp := os.Getenv("SOURCE_AUTH_PASSWORD"); sp != "" {
+			sourceAuth.Password = sp
+		}
+		config.Source.Repo.Auth = sourceAuth
```
jotadrilo, line 40: A quick comment that apply to the other block:

Use short variables when it makes sense (like here). The [basic rule](https://github.com/golang/go/wiki/CodeReviewComments#variable-names): the further from its declaration that a name is used, the more descriptive the name must be.

Hence, you can use `u` and `p` to refer to `username` and `password` in the `if-Getenv` scope and `a` (or `auth`) to refer to `api.Auth` in the `if-Auth` scope.

Another approach:

```suggestion
		config.Source.Repo.Auth := &api.Auth{}
		if u := os.Getenv("SOURCE_AUTH_USERNAME"); u != "" {
			config.Source.Repo.Auth.Username = u
		}
		if p := os.Getenv("SOURCE_AUTH_PASSWORD"); p != "" {
			config.Source.Repo.Auth.Password = p
		}
```

Another point is, does it make sense to provide username without a password? Otherwise, I would do a double-check:

```suggestion
		if u, p := os.Getenv("SOURCE_AUTH_USERNAME"), os.Getenv("SOURCE_AUTH_PASSWORD"); u != "" && p != "" {
			config.Source.Repo.Auth = &api.Auth{Username: u, Password: p}
        }
```

And finally, you could even add a method to the `api` package to set the password, so you don't have to duplicate it:

```
# api.go

func (r *Repo) SetBasicAuth(user string, pass string) error {
  // No auth provided, leave as-is
  if user == "" && pass == "" {
      return nil
  }
  // TODO: must both be set? then manage each case

  r.Auth = &Auth{Username: user, Password: pass}
  return nil
}
```

so you can simplify the code to:

```
	if config.Target.Repo.Auth == nil {
        if err := config.Target.Repo.SetBasicAuth(os.Getenv("TARGET_AUTH_USERNAME"), os.Getenv("TARGET_AUTH_PASSWORD")); err != nil {
            return err
        }
    }
```

However, I am just mentioning these options to you. Another option is to leave the code as-is right now if you prefer.

pkg/config/config_test.go
```diff
@@ -34,3 +35,35 @@ func TestLoad(t *testing.T) {
...
+	if source.Repo.Auth.Password != "sPassword" {
+		t.Errorf("Got: %s, want %s", source.Repo.Auth.Password, "sPassword")
+	}
+	if target.Repo.Auth.Username != "tUsername" {
+		t.Errorf("Got: %s, want %s", target.Repo.Auth.Username, "tUsername")
+	}
+	if target.Repo.Auth.Password != "tPassword" {
+		t.Errorf("Got: %s, want %s", target.Repo.Auth.Password, "tPassword")
+	}
+}
```
jotadrilo, line 69: Please add a test case table with different use cases. For example, providing the username in the YAML but the password from env vars, providing auth in the YAML and empty (but existing) env vars, etc.

tompizmor:
Thanks for the suggestions @jotadrilo! 

I moved the code to an API function and adds several test cases to check different combinations.

pkg/config/config_test.go
```diff
@@ -38,32 +39,112 @@ func TestLoad(t *testing.T) {
...
-	// If a config file is found, read it in.
-	if err := viper.ReadInConfig(); err != nil {
-		t.Fatalf("error reading config file: %+v", err)
-	}
-	if err := Load(&syncConfig); err != nil {
-		t.Fatalf("error loading config file")
+	tests := map[string]struct {
+		inputFile string
+		envVars   map[string]string
+		expected  map[string]string
```
jotadrilo, line 45: You should compare the final output, not part of it separately. Otherwise, you may introduce bugs in other fields (if you extended the object, for example) and don't notice a side effect. In this case, the `api.Auth` object. This way you need to compare with `reflect` or `proto` packages:

```
expectedSourceAuth *api.Auth
expectedTargetAuth *api.Auth

...

			"example-config-no-auth.yaml",
			map[string]string{
				"SOURCE_AUTH_USERNAME": "sUsername",
				"SOURCE_AUTH_PASSWORD": "sPassword",
				"TARGET_AUTH_USERNAME": "tUsername",
				"TARGET_AUTH_PASSWORD": "tPassword",
			},
			&api.Auth{Username: "sUsername", Password: "sPassword"},
			&api.Auth{Username: "tUsername", Password: "tPassword"},

...

			if got, want := source.Repo.Auth, tc.expectedSourceAuth; !proto.Equal(got, want) {
				t.Errorf("got: %+v; want %+v", got, want)
			}
```

Note that the pattern in Go is to use `got` and `want` in tests.

api/api.go
```diff
@@ -2,3 +2,24 @@
...
+	if r.Auth == nil {
+		r.Auth = &Auth{Username: username, Password: password}
+	} else {
+		if r.Auth.Username == "" {
+			r.Auth.Username = username
+		}
+		if r.Auth.Password == "" {
+			r.Auth.Password = password
+		}
+	}
```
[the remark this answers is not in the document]
tompizmor: Yes, I already changed that. Probably last time you checked it was an old version

pkg/config/config_test.go
tompizmor: done!

PR #35, opened 2020-08-06
Avoid download target index.yaml several times

### LINKED ISSUE #33: Execution lasts a long time when using a target HTTPS repo
It seems currently the tool is downloading the `index.yaml` per chart and version, what causes execution to be delayed.

```
I0803 13:28:18.365226   21482 harbor.go:49] Checking if zookeeper-5.21.1 chart exists in "https://harbor-foo.bar"
I0803 13:28:18.365238   21482 utils.go:71] Repo configures basic authentication. Downloading index.yaml...
...
I0803 13:28:33.054737   21482 harbor.go:49] Checking if zookeeper-5.21.0 chart exists in "https://harbor-foo.bar"
I0803 13:28:33.054748   21482 utils.go:71] Repo configures basic authentication. Downloading index.yaml...
```
in the above example, first is looking for `zookeeper-5.21.1` where the index.yaml is downloaded, once checked this version, the tool is going to check `zookeeper-5.21.0` and the index.yaml is downloaded again.

When using HTTPS and basic authentication the download of the index.yaml is not instantaneous, taking between 15-20 seconds per chart and version. That means when synchronizing a repository with a huge number of charts/versions the execution is going to take a lot of time:
```
I0803 13:28:33.054737   21482 harbor.go:49] Checking if zookeeper-5.21.0 chart exists in "https://harbor-foo.bar"
I0803 13:28:33.054748   21482 utils.go:71] Repo configures basic authentication. Downloading index.yaml...
I0803 13:28:51.806207   21482 utils.go:41] Chart "zookeeper" exists in index.yaml file. Searching "5.21.0" version
I0803 13:28:51.806226   21482 utils.go:44] Version "5.21.0" found for chart "zookeeper" in index.yaml file
```
18 seconds in the above example, from `13:28:33` to `13:28:51` according to the timestamp.

In my current use case, just synchronizing a catalog of 39 charts with versions from the last 20 days, it is taking around 15 minutes without any push as all the charts are up-to-date.

tompizmor:
This PR is a small refactor so that the `chartsExists()` function does not download the target index.yaml in each call. We detect this was making the tool to run slowly in certain environments.

Additionally, it fixes an issue with the `FindChartURL()` function when the URL stored in the index.yaml is relative and not absolute.

Fixes #33

### DIFF
--- a/pkg/chartrepotest/chartmuseumreal.go
+++ b/pkg/chartrepotest/chartmuseumreal.go
@@ -45,7 +45,7 @@ func tChartMuseumReal(t *testing.T, username, password string) (string, func()) 
 		id,
 	))
 
-	return "http://localhost:" + port, func() {
+	return "http://127.0.0.1:" + port, func() {
 		tDockerExec(t, "stop", id)
 		tDockerExec(t, "rm", id)
 		tDockerVolumeRm(t, chartsVolume)
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"os"
 	"os/exec"
 	"time"
@@ -136,10 +137,14 @@ func GetDateThreshold(date string) (time.Time, error) {
 }
 
 // FindChartURL will return the chart url
-func FindChartURL(name string, version string, index *helmRepo.IndexFile) (string, error) {
+func FindChartURL(name string, version string, index *helmRepo.IndexFile, sourceUrl string) (string, error) {
 	chart := findChartByVersion(index.Entries[name], version)
 	if chart != nil {
-		return chart.URLs[0], nil
+		if isValidURL(chart.URLs[0]) {
+			klog.Infof("%s is valid url", chart.URLs[0])
+			return chart.URLs[0], nil
+		}
+		return fmt.Sprintf("%s/%s", sourceUrl, chart.URLs[0]), nil
 	}
 	return "", fmt.Errorf("unable to find chart url in index")
 }
@@ -152,4 +157,17 @@ func findChartByVersion(chartVersions []*helmRepo.ChartVersion, version string) 
 		}
 	}
 	return nil
+}
+
+// isValidUrl tests a string to determine if it is a well-structured url or not.
+func isValidURL(text string) bool {
+	_, err := url.ParseRequestURI(text)
+	if err != nil {
+		return false
+	}
+	u, err := url.Parse(text)
+	if err != nil || u.Scheme == "" || u.Host == "" {
+		return false
+	}
+	return true
 }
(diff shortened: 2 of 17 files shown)


pkg/chartrepotest/chartmuseumreal.go
```diff
@@ -45,7 +45,7 @@ func tChartMuseumReal(t *testing.T, username, password string) (string, func())
 		id,
 	))
 
-	return "http://localhost:" + port, func() {
+	return "http://127.0.0.1:" + port, func() {
```
aruiz14, line 48: Out of curiosity, why did you need this?

pkg/utils/utils.go
```diff
@@ -136,10 +137,13 @@ func GetDateThreshold(date string) (time.Time, error) {
 }
 
 // FindChartURL will return the chart url
-func FindChartURL(name string, version string, index *helmRepo.IndexFile) (string, error) {
+func FindChartURL(name string, version string, index *helmRepo.IndexFile, sourceUrl string) (string, error) {
 	chart := findChartByVersion(index.Entries[name], version)
 	if chart != nil {
```
aruiz14, line 142: You should probably also check if the `URLs` array is not empty to avoid an out-of-index error.

aruiz14:
```suggestion
	if chart != nil && len(chart.URLs) > 0 {
```

pkg/utils/utils.go
```diff
@@ -153,3 +157,16 @@ func findChartByVersion(chartVersions []*helmRepo.ChartVersion, version string)
...
 	return nil
 }
+
+// isValidUrl tests a string to determine if it is a well-structured url or not.
+func isValidURL(text string) bool {
+	_, err := url.ParseRequestURI(text)
+	if err != nil {
+		return false
+	}
+	u, err := url.Parse(text)
```
aruiz14, line 167: Is it really needed to parse it twice? Here you can find how both `Parse` and `ParseRequestURI` basically share the same parsing method:
 - https://golang.org/src/net/url/url.go?s=14096:14145#L485
 - https://golang.org/src/net/url/url.go?s=13410:13449#L464

If it does matter, please add a comment.

pkg/chartrepotest/chartmuseumreal.go
tompizmor: I had some tests failing and I thought this was related. At the end, it was another thing but I kept the change.

pkg/utils/utils.go
tompizmor: good point

pkg/utils/utils.go
tompizmor: Probably not needed. I took the function from somewhere else and use it as it was.

tompizmor:
I tested this version in the scenario where we identified the slowness and the same command execution time decreased from 32min to only 40 sec.

PR #36, opened 2020-09-29
Add support for Helm v3 dependency management

tompizmor:
Since Helm v3 the recommended way to specify chart dependencies is directly in the Chart.yaml file and Chart.lock instead of requirements.yaml and requirements.lock. 

This PR adds support for both kinds of dependency management by checking the existence of `requirements.lock` or `Chart.lock` in the processed chart.


pkg/chart/dependency.go
```diff
@@ -22,20 +22,28 @@ type dependencies struct {
...
-	requirementsLockFile := path.Join(chartPath, "requirements.lock")
+	lockFileName := ""
+	if apiVersion == "v1" {
+		lockFileName = "requirements.lock"
+	} else if apiVersion == "v2" {
+		lockFileName = "Chart.lock"
+	} else {
+		return errors.Errorf("unrecognised apiVersion %s", apiVersion)
+	}
+	lockFile := path.Join(chartPath, lockFileName)
```
jotadrilo, line 37: I would add a small function for this (and a test separately)

```
func lockFilePath(chartPath, apiVersion) (string, error) {
}
```

```
err, lockFile := lockFilePath(chartPath, apiVersion)
if err != nil { ... }
```

pkg/chart/dependency.go
```diff
@@ -22,20 +22,28 @@ type dependencies struct {
...
 	var missingDependencies = false
-	requirementsLockFile := path.Join(chartPath, "requirements.lock")
+	lockFileName := ""
+	if apiVersion == "v1" {
+		lockFileName = "requirements.lock"
+	} else if apiVersion == "v2" {
+		lockFileName = "Chart.lock"
+	} else {
+		return errors.Errorf("unrecognised apiVersion %s", apiVersion)
+	}
```
jotadrilo, line 36: please be consistent with the usage of `if` or `switch`

pkg/chart/dependency.go
```diff
@@ -77,9 +85,18 @@ func syncDependencies(chartPath string, sourceRepo *api.Repo, target *api.Target
 	}
 
 	if !missingDependencies {
-		klog.V(3).Info("Updating requirements.yaml file...")
-		if err := updateRequirementsFile(chartPath, lock, sourceRepo, target); err != nil {
-			return errors.Trace(err)
+		klog.V(3).Info("Updating Chart.yaml file...")
+		switch apiVersion {
+		case "v1":
```
jotadrilo, line 90: I would add a `types.go` file stating this kind of constants:

```golang
package chart

const (
  API_V1 string = "v1"
  API_V2 string = "v2"
)
```

you can use this file to define common constants such as "Chart.yaml"

pkg/chart/dependency.go
tompizmor: done!

pkg/chart/dependency_test.go
```diff
@@ -163,3 +221,51 @@ func TestFindDepByName(t *testing.T) {
...
+			errors.New("unrecognised apiVersion \"vvv000\""),
+		},
+	}
+
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			want := tc.expectedPath
+			got, err := lockFilePath(tc.chartPath, tc.apiVersion)
+			if tc.shouldFail {
+				if !assert.EqualError(t, tc.expectedError, err.Error()) {
```
jotadrilo, line 261: why do you need `assert`?

pkg/chart/dependency.go
```diff
@@ -77,9 +80,18 @@ func syncDependencies(chartPath string, sourceRepo *api.Repo, target *api.Target
 	}
 
 	if !missingDependencies {
-		klog.V(3).Info("Updating requirements.yaml file...")
-		if err := updateRequirementsFile(chartPath, lock, sourceRepo, target); err != nil {
-			return errors.Trace(err)
+		klog.V(3).Info("Updating Chart.yaml file...")
```
jotadrilo, line 83: I would just say "Updating dependencies file..."

pkg/chart/dependency.go
```diff
@@ -88,9 +100,39 @@ func syncDependencies(chartPath string, sourceRepo *api.Repo, target *api.Target
...
+func updateChartMetadataFile(chartPath string, lock *helmChart.Lock, sourceRepo *api.Repo, target *api.TargetRepo) error {
+	chartFile := path.Join(chartPath, ChartFilename)
+	chart, err := ioutil.ReadFile(chartFile)
+	if err != nil {
+		return errors.Trace(err)
+	}
+	chartMetadata := &helmChart.Metadata{}
+	err = yaml.Unmarshal(chart, chartMetadata)
+	if err != nil {
+		return errors.Annotatef(err, "Error unmarshaling %s file", chartFile)
```
jotadrilo, line 114: remember that errors should not be capitalized: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

jotadrilo (approved):
LGTM!

pkg/chart/dependency_test.go
tompizmor: I don't really need it. I can compare errors strings:

~~~go
err.Error() != tc.expectedError.Error() {
...
}
~~~

pkg/chart/dependency.go
tompizmor: Just for you to know, the new errors I wrote in this PR I did it in lower case, but this one comes from a copy-paste.

Hopefully, you don't have to remind me this rule anymore 😄

PR #37, opened 2020-09-30
Lowercase all errors

tompizmor:
As recommended in the Golang code review conventions, errors string should not be capitalized.

https://github.com/golang/go/wiki/CodeReviewComments#error-strings


jotadrilo (approved):
LGTM!

PR #47, opened 2020-11-18
feat(syncer): Add syncer pkg

jotadrilo:
This patch adds a `syncer` package to encapsulate the sync logic.

This initial encapsulation is pretty basic. If we plan to add support for other backends, it should be refactored further to decouple the sync logic from the backend implementation while keeping a clean interface.


tompizmor (approved):
lgtm

PR #48, opened 2020-11-18
feat(api): Add charts array to the configuration

jotadrilo:
This patch only adds the `charts` array to the configuration schema.


tompizmor (approved):
lgtm

PR #51, opened 2020-11-20
fix(pkg): Refactor chart clients

jotadrilo:
This patch refactor the chart clients to implement a common interface not tied to the technical implementation.

Also, each client ties to itself, so it does not require providing an `api.Repo` to its methods.

### DIFF
--- a/pkg/client/core/client.go
+++ b/pkg/client/core/client.go
@@ -1,4 +1,4 @@
-package repo
+package core
 
 import (
 	"fmt"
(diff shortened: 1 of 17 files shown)


pkg/client/core/client.go
```diff
@@ -1,24 +1,33 @@
...
+// Reader defines the methods that a ReadOnly chart client should implement.
+type Reader interface {
+	Fetch(filepath string, name string, version string) error
+	ChartExists(name string, version string) (bool, error)
+}
+
+// Writer defines the methods that a WriteOnly chart client should implement.
+type Writer interface {
+	Push(filepath string) error
 }
```
jotadrilo, line 24: These are the most important changes

jotadrilo:
Closing in favor of #54

PR #54, opened 2020-11-23
feat(core): Add currently supported clients

jotadrilo:
This patch adds all client implementations under `pkg/core`. It also adds the new `Reader` and `Writer` interfaces that will allow us to make `Client` interface to rely on them directly.

This rework puts implementation details into each client, which will allow us to add new clients without altering the API nor existing clients.


jotadrilo:
Closing in favor of #54

pkg/client/core/core.go
```diff
@@ -0,0 +1,59 @@
...
+	Fetch(filepath string, name string, version string) error
+	List() ([]string, error)
+	ListChartVersions(name string) ([]string, error)
+	Has(name string, version string) (bool, error)
+}
+
+// Writer defines the methods that a WriteOnly chart client should implement.
+type Writer interface {
+	Upload(filepath string) error
+}
```
jotadrilo, line 24: these are the new interfaces

tompizmor (approved):
LGTM!

PR #56, opened 2020-11-23
feat(syncer): Enhance syncer struct

jotadrilo:
This patch adds a `New` method, which will return an enhanced version of `Syncer` containing the new `ClientV2` chart repo clients.

These clients will replace the current ones as we implement pending features.

### DIFF
--- a/pkg/syncer/syncer.go
+++ b/pkg/syncer/syncer.go
@@ -14,10 +14,18 @@ import (
 	"github.com/bitnami-labs/charts-syncer/pkg/utils"
 )
 
+// Clients holds the source and target chart repo clients
+type Clients struct {
+	src *core.ClientV2
+	dst *core.ClientV2
+}
+
 // A Syncer can be used to sync a source and target chart repos.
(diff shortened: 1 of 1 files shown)


pkg/syncer/syncer.go
```diff
@@ -14,11 +14,19 @@ import (
 	"github.com/bitnami-labs/charts-syncer/pkg/utils"
 )
 
+// Clients holds the source and target chart repo clients
+type Clients struct {
+	src core.ClientV2
+	dst core.ClientV2
```
tompizmor, line 20: I would try to avoid the word `destination` in favor of `target`, to be consistent with the config file and other variables/functions

jotadrilo: can we refactor this wording after I merge all the PRs? Otherwise, I need to rebase and fix the #57, #58 and #59

tompizmor: makes sense yes!

tompizmor (approved):

PR #57, opened 2020-11-23
feat(client): Add Reload method

jotadrilo:
To reload the `index.yaml` file is an expensive task. This patch implements the `Reload` method in the `Client` API so processes will call it on-demand to refresh chart repo clients.

### DIFF
--- a/pkg/client/chartmuseum/chartmuseum.go
+++ b/pkg/client/chartmuseum/chartmuseum.go
@@ -39,13 +39,16 @@ func New(repo *api.Repo) (*Repo, error) {
 		return nil, errors.Trace(err)
 	}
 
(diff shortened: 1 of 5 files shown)


tompizmor:
is it expensive because it has to download the index from the target repo?

jotadrilo:
> is it expensive because it has to download the index from the target repo?

The expensive task is the GET request to any repo storing an `index.html` file. It delays the process a few seconds. If we need to sync let's say 1000 charts (reality is more than that) we cannot reload the `index.html` file 1000 times (if each GET takes 1s, that's >15min).

Therefore, this path delegates the `Reload` task to the client consumer.

tompizmor (approved):

PR #58, opened 2020-11-23
fix(syncer): Split code into different files

jotadrilo:
This patch splits `syncer.go` in different files to ease future features.


tompizmor (approved):
LGTM

PR #59, opened 2020-11-23
feat(syncer): Add topological sort for charts

jotadrilo:
This patch adds [topological sort](https://en.wikipedia.org/wiki/Topological_sorting) capabilities to the `Syncer`.

This allows the `Syncer` to compute the list of charts to synchronize in a way that eases the subsequent uploads to the target chart repository.

### DIFF
--- a/pkg/chart/dependency.go
+++ b/pkg/chart/dependency.go
@@ -3,17 +3,20 @@ package chart
 import (
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path"
 
-	"github.com/bitnami-labs/charts-syncer/api"
-	"github.com/bitnami-labs/charts-syncer/pkg/client/core"
-	"github.com/bitnami-labs/charts-syncer/pkg/helmcli"
 	"github.com/juju/errors"
 	"github.com/mkmik/multierror"
 	helmChart "helm.sh/helm/v3/pkg/chart"
 	helmRepo "helm.sh/helm/v3/pkg/repo"
 	"k8s.io/klog"
 	"sigs.k8s.io/yaml"
+
+	"github.com/bitnami-labs/charts-syncer/api"
+	"github.com/bitnami-labs/charts-syncer/pkg/client/core"
+	"github.com/bitnami-labs/charts-syncer/pkg/helmcli"
+	"github.com/bitnami-labs/charts-syncer/pkg/utils"
 )
 
 // dependencies is the list of dependencies of a chart
@@ -97,7 +100,7 @@ func syncDependencies(chartPath string, sourceRepo *api.Repo, target *api.Target
 			return errors.Trace(err)
 		}
 	}
-	return errs
+	return errors.Trace(errs)
 }
 
 // updateChartMetadataFile updates the dependencies in Chart.yaml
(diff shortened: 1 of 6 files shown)


pkg/chart/dependency.go
```diff
@@ -204,3 +207,46 @@ func lockFilePath(chartPath, apiVersion string) (string, error) {
...
+	var apiVersion string
+	if _, err := os.Stat(path.Join(chartPath, RequirementsLockFilename)); err == nil {
+		apiVersion = APIV1
+	}
+	if _, err := os.Stat(path.Join(chartPath, ChartLockFilename)); err == nil {
+		apiVersion = APIV2
+	}
+
+	if apiVersion == "" {
+		return nil, nil
```
tompizmor, line 236: shouldn't we return an error in this case?

pkg/syncer/index.go
```diff
@@ -0,0 +1,195 @@
...
+		}
+	}
+
+	return errors.Trace(errs)
+}
+
+// loadChart loads a chart in the chart index map
+func (s *Syncer) loadChart(name string, version string) error {
+	id := fmt.Sprintf("%s-%s", name, version)
+	if ch := s.getIndex().Get(id); ch != nil {
```
tompizmor, line 127: aren't you checking this already before calling the function? 

Line 109 of this same file

pkg/syncer/sync.go
```diff
@@ -88,3 +88,31 @@ func (s *Syncer) Sync(charts ...string) error {
...
+	}
+	charts, err := s.topologicalSortCharts()
+	if err != nil {
+		return errors.Trace(err)
+	}
+
+	var errs error
+	for _, chart := range charts {
+		id := fmt.Sprintf("%s-%s", chart.Name, chart.Version)
+		// 1. Process tgz file
```
tompizmor, line 107: Is this WIP yet?

pkg/chart/dependency.go
jotadrilo: I think this is a bit unclear.

We only set `apiVersion` if any of the possible "lock" files are found. There are charts without a "lock" file so this `if` clause is covering that case. Should I add a comment?

pkg/syncer/sync.go
jotadrilo: yes! Do you prefer to take this method off of the PR?

pkg/syncer/index.go
jotadrilo: yes! The difference is that `loadChart` is a recursive function and it will be invoked again for each dependency. It makes sense that different "tier1" charts use the same "tier2" chart dependencies. This check will make the method to skip already indexed charts.

Should I clarify it with a comment?

tompizmor: yes, please!

pkg/chart/dependency.go
tompizmor: I see...I thought you were only calling this function IF the chart has dependencies but it's been called for all charts. In that case yes, it makes sense that we return nil with no error.

What I find wrong is that we are using apiVersion for that. In order to check the API, we should be checking the Chart.yaml file and additionally check if there are deps or not.

jotadrilo: I see your point. However, that also depends on the API version. I guess you were doing this quick check to identify the API version right? I assume all charts are well-formed so we don't need to validate if we miss the dependencies in the "main" file.

tompizmor: Yes, it depends on the API version, but we don't know if charts are really well-formed or not. Maybe someone in one repo has pushed a V1 chart with a Chart.lock file (which is from v2). In that case, we would think the chart is V2. Although probably that wrong chart will fail later because of another reason.

In any case, I was just stating that as a future improvement, I know it was already like that.

tompizmor (approved):

pkg/chart/dependency.go
jotadrilo: I think it is not the responsibility of this tool to manage that case. If that ever happens, any tool consuming that chart would fail, not only this syncer.

PR #60, opened 2020-11-24
feat(syncer): Add SyncPendingCharts method

jotadrilo:
This patch adds a hidden command to allow us to develop and test the experimental clients.

### DIFF
--- a/cmd/syncpending.go
+++ b/cmd/syncpending.go
@@ -28,6 +28,7 @@ func newSyncPendingCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "sync-pending",
 		Short:   "[EXPERIMENTAL] Synchronizes two chart repositories using an experimental feature",
+		Hidden:  true,
 		Example: syncPendingExample,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			if err := initConfigFile(); err != nil {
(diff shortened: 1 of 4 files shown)


cmd/syncpending.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+var (
+	syncPendingFromDate string
+)
+
+var (
+	syncPendingExample = `
+  # Synchronizes all charts defined in the configuration file
+  charts-syncer sync
```
tompizmor, line 19: `sync` -> `sync-pending`

tompizmor (commented):
What's the difference with the current `sync` command?

The `sync` one iterates over dependencies as they are found and this one uses the topological sort first?

cmd/syncpending.go
jotadrilo: > What's the difference with the current `sync` command?
> 
> The `sync` one iterates over dependencies as they are found and this one uses the topological sort first?

I am just leaving `sync` intact while I work on the new sync under `sync-pending`. Once all of this is done, we will only have `sync`.

In any case, yes, this new method takes care of getting a list of charts topologically sorted and sync then without worrying about dependencies anymore.

tompizmor (approved):

PR #61, opened 2020-11-25
feat(syncer): Add SyncPendingCharts method

jotadrilo:
The current `Syncer` stores charts in a temporary folder. This patch adds a local working directory to prevent downloading the charts over and over in each run.

This is not especially useful for a real case usage of this tool in which we will only download a big number of charts the first time syncing two chart repos. We can consider that subsequent runs will download only a few of them. In any case, subcharts (dependencies) might be the same in these subsequent runs, so it still makes sense to add this feature to improve the performance of the tool in real cases.

The reality is that this is extremely useful to speed up the development process.

Example:

```shell
❯ echo -n https://charts.bitnami.com/bitnami | sha1sum -t
cc54f5167101a3913ef8e9c2aff515129771b204  -

❯ tree ~/.charts-syncer/
/Users/jdrios/.charts-syncer/
└── cc54f5167101a3913ef8e9c2aff515129771b204
    ├── common-0.7.1.tgz
    ├── common-0.8.0.tgz
    ├── common-1.0.0.tgz
    ├── common-1.0.1.tgz
    ├── ghost-10.1.15.tgz
    ├── ghost-10.1.16.tgz
    ├── ghost-10.1.17.tgz
    ├── ghost-10.1.18.tgz
    ├── ghost-10.1.19.tgz
    ├── ghost-10.1.20.tgz
    ├── ghost-10.2.0.tgz
    ├── ghost-10.2.1.tgz
    ├── ghost-10.2.2.tgz
    ├── ghost-10.2.3.tgz
    ├── ghost-11.0.0.tgz
    ├── mariadb-7.10.3.tgz
    ├── mariadb-7.10.4.tgz
    └── mariadb-9.0.1.tgz

1 directory, 18 files
```

### DIFF
--- a/pkg/syncer/index.go
+++ b/pkg/syncer/index.go
@@ -1,9 +1,8 @@
 package syncer
 
 import (
+	"crypto/sha1"
 	"fmt"
-	"io/ioutil"
-	"os"
 	"path"
 	"sort"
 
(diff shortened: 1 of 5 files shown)


pkg/syncer/index.go
```diff
@@ -148,16 +147,15 @@ func (s *Syncer) loadChart(name string, version string) error {
...
-	workdir, err := ioutil.TempDir("", "charts-syncer")
-	if err != nil {
-		return errors.Trace(err)
-	}
-	defer os.RemoveAll(workdir)
+	tgz := path.Join(s.srcWorkdir, fmt.Sprintf("%s-%s.tgz", name, version))
 
-	tgz := path.Join(workdir, fmt.Sprintf("%s-%s.tgz", name, version))
-	if err := s.cli.src.Fetch(tgz, name, version); err != nil {
+	// Fetch chart iff it does not exists in the workdir already
```
tompizmor, line 152: `iff` -> `if`
`exists` -> `exist`

tompizmor (approved):

PR #64, opened 2020-11-25
feat(client): Improve HTTP response handling

jotadrilo:
The current client implementations are only taking 404 HTTP responses into account. However, we might receive other non-2xx errors.

This patch will make clients take into account all non-2xx responses, instead. It also adds an string identifier to each query log.


tompizmor (approved):
lgtm

PR #65, opened 2020-11-25
fix(misc): Improve some log messages

jotadrilo:
This patch improves some log messages only.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -25,9 +25,9 @@ func Load(config *api.Config) error {
 	if err != nil {
 		return errors.Trace(fmt.Errorf("error unmarshalling config file: %w", err))
 	}
-	if config.Target.RepoName == "" {
-		klog.Warning("'target.repoName' property is empty. Using 'myrepo' default value")
-		config.Target.RepoName = defaultRepoName
+	if config.GetTarget().GetRepoName() == "" {
+		klog.V(4).Infof("'target.repoName' property is empty. Using %q default value", defaultRepoName)
+		config.GetTarget().RepoName = defaultRepoName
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -25,9 +25,9 @@ func Load(config *api.Config) error {
 	if err != nil {
 		return errors.Trace(fmt.Errorf("error unmarshalling config file: %w", err))
 	}
-	if config.Target.RepoName == "" {
-		klog.Warning("'target.repoName' property is empty. Using 'myrepo' default value")
-		config.Target.RepoName = defaultRepoName
+	if config.GetTarget().GetRepoName() == "" {
```
tompizmor, line 28: just out of curiosity, why do you prefer this way?

tompizmor (approved):

pkg/config/config.go
jotadrilo: It is not about my preference. When we use protocol buffer the Go compiles generates a `pb.go` file containing all the required types (e.g. `struct`s, `map`s, etc) with named fields (e.g. `Target`) and accessor methods (e.g. `GetTarget()`) which returns the values set (or the default value if the field is unset). If the default is not explicitly set, the zero value of that type is used instead.

This is especially useful when you are accessing fields inside structs, since the accessors will manage the zero value properly too. For example, if `Target` is `nil`, `Target.GetRepoName()` would work (since `GetRepoName()` is a receiver function of `Target`, regardless its value), but `Target.RepoName` would fail (since `nil.RepoName` is inaccessible).

tompizmor: 👍

PR #67, opened 2020-12-01
feat(client): Add support for OCI registries

tompizmor:
This PR is still in progress.

It already supports syncing charts from OCI registries but not to them.

- [x] Currently working on building dependencies from the target OCI repository.

In order to push charts, we are using now helm CLI although it is not the ideal implementation.


pkg/chart/sync.go
```diff
@@ -165,8 +165,29 @@ func ChangeReferences(outDir, filepath, name, version string, srcRepo *api.Sourc
 	// does not take into account that the target repo could be out of sync yet
 	// (for example, if we uploaded a dependency of the chart being packaged a
 	// few seconds ago).
-	if err := helmcli.UpdateDependencies(chartPath); err != nil {
-		return "", errors.Trace(err)
+	if tgtRepo.Repo.Kind == api.Kind_OCI {
```
jotadrilo, line 168: let's use a `switch`, instead

pkg/chart/sync.go
```diff
@@ -165,8 +165,29 @@ func ChangeReferences(outDir, filepath, name, version string, srcRepo *api.Sourc
...
+		// Build deps manually for OCI as helm does not supported it yet
```
jotadrilo, line 169:
```suggestion
		// Build deps manually for OCI as helm does not support it yet
```

pkg/chart/sync.go
jotadrilo, line 168:
```suggestion
	if tgtRepo.GetRepo().GetKind() == api.Kind_OCI {
```

pkg/client/chartmuseum/chartmuseum.go
```diff
@@ -53,7 +53,7 @@ func (r *Repo) GetUploadURL() string {
 }
 
 // Upload uploads a chart to the repo.
-func (r *Repo) Upload(filepath string) error {
+func (r *Repo) Upload(filepath, name, version string) error {
```
jotadrilo, line 56:
```suggestion
func (r *Repo) Upload(filepath, _, _ string) error {
```

pkg/client/harbor/harbor.go
```diff
@@ -54,7 +54,7 @@ func (r *Repo) GetUploadURL() string {
 }
 
 // Upload uploads a chart to the repo
-func (r *Repo) Upload(filepath string) error {
+func (r *Repo) Upload(filepath, name, version string) error {
```
jotadrilo, line 57:
```suggestion
func (r *Repo) Upload(filepath, _, _ string) error {
```

pkg/client/helmclassic/helmclassic.go
```diff
@@ -198,7 +198,7 @@ func (r *Repo) Has(name string, version string) (bool, error) {
 }
 
 // Upload uploads a chart to the repo
-func (r *Repo) Upload(filepath string) error {
+func (r *Repo) Upload(filepath, name, version string) error {
```
jotadrilo, line 201:
```suggestion
func (r *Repo) Upload(_, _, _ string) error {
```

tompizmor: done

pkg/client/harbor/harbor.go
tompizmor: done

pkg/client/chartmuseum/chartmuseum.go
tompizmor: done

pkg/chart/sync.go
tompizmor: done

pkg/chart/sync.go
tompizmor: done

pkg/chart/sync.go
tompizmor: done

pkg/syncer/sync.go
```diff
@@ -145,20 +160,75 @@ func (s *Syncer) SyncPendingCharts(names ...string) error {
...
+			if err := chart.ChangeDependenciesFile(chartPath, ch.Name, s.source, s.target); err != nil {
+				return errors.Trace(err)
+			}
+			switch s.target.GetRepo().GetKind() {
+			case api.Kind_HELM:
+			case api.Kind_CHARTMUSEUM:
+			case api.Kind_HARBOR:
+				if err := helmcli.UpdateDependencies(chartPath); err != nil {
+					return errors.Trace(err)
+				}
```
jotadrilo: please put these under the `default` case

pkg/syncer/sync.go
```diff
@@ -145,20 +160,75 @@ func (s *Syncer) SyncPendingCharts(names ...string) error {
...
+				if err != nil {
+					return errors.Trace(err)
+				}
+				// Dependencies found
+				if lock != nil {
+					for _, dep := range lock.Dependencies {
+						depTgz := path.Join(chartPath, "charts", fmt.Sprintf("%s-%s.tgz", dep.Name, dep.Version))
+						s.cli.dst.Fetch(depTgz, dep.Name, dep.Version)
+					}
+				}
```
jotadrilo, line 204: please encapsulate this task into a function, it deserves proper unit testing in an isolated way and the code will look simpler

tompizmor: done!

api/config.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.23.0
+// 	protoc-gen-go v1.25.0-devel
```
jotadrilo, line 3: We should not use a `devel` version here, I guess. Also, we should add some kind of automation to prevent us from using different `protoc-gen-go` versions.

pkg/client/oci/oci.go
```diff
@@ -0,0 +1,302 @@
...
+	Digest    string
+	Size      int
+}
+
+// TagManifest contains info about the published TAG in OCI
+type TagManifest struct {
+	SchemaVersion int
+	Config        LayerInfo
+	Layers        []LayerInfo
+}
```
jotadrilo, line 58: have you checked the [upstream OCI schema](https://godoc.org/github.com/moby/moby/vendor/github.com/docker/distribution/manifest/ocischema#Manifest)?

pkg/client/oci/oci.go
```diff
@@ -0,0 +1,302 @@
...
+
+// KnownMediaTypes returns a list of layer mediaTypes that the Helm client knows about
+func KnownMediaTypes() []string {
+	return []string{
+		HelmChartConfigMediaType,
+		HelmChartContentLayerMediaType,
+	}
+}
```
jotadrilo, line 66: check [`ManifestMediaTypes`](https://godoc.org/github.com/docker/distribution#ManifestMediaTypes)

pkg/client/oci/oci.go
```diff
@@ -0,0 +1,302 @@
...
+	if err != nil {
+		return nil, errors.Errorf("unexpected response — %d %q — from %s", status, http.StatusText(status), u.String())
+	}
+	// If status not found 404, it could mean the asset has no release yet
+	if status == http.StatusNotFound {
+		return []string{}, nil
+	}
+	if status != http.StatusOK {
+		return nil, errors.Errorf("unexpected response — %d %q, %s — from %s", status, http.StatusText(status), string(body), u.String())
+	}
```
jotadrilo, line 179: let's use a `switch`. Are these response codes oci-specific or are they harbor-oci-specific?

pkg/client/oci/oci.go
```diff
@@ -0,0 +1,302 @@
...
+	if err != nil {
+		return errors.Annotatef(err, "fetching %s:%s chart", name, version)
+	}
+	defer res.Body.Close()
+
+	if ok := res.StatusCode >= 200 && res.StatusCode <= 299; !ok {
+		bodyStr := utils.HTTPResponseBody(res)
+		return errors.Errorf("unable to fetch %s:%s chart, got HTTP Status: %s, Resp: %v", name, version, res.Status, bodyStr)
+	}
+	klog.V(4).Infof("[%s] Got HTTP Status: %s", reqID, res.Status)
```
jotadrilo, line 240: let's use a `switch` and check the specific response codes that the API may return

pkg/client/oci/oci.go
```diff
@@ -0,0 +1,302 @@
...
+		}
+	}
+	return chartTags, nil
+}
+
+// GetDownloadURL returns the URL to download a chart
+func (r *Repo) GetDownloadURL(name string, version string) (string, error) {
+	digest, err := r.getChartDigest(name, version)
+	if err != nil {
+		return "", errors.Trace(err)
```
jotadrilo, line 204:
```suggestion
		return "", errors.Annotatef(err, "obtaining chart %s:%s digest")
```

pkg/client/oci/oci.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+	for _, layer := range tm.Layers {
+		if layer.MediaType == "application/tar+gzip" {
+			return layer.Digest, nil
+		}
+	}
+
+	return "", errors.Errorf("unable to found digest for %s:%s", name, version)
```
jotadrilo, line 142:
```suggestion
	return "", errors.NotFoundf("digest")
```

pkg/client/oci/oci.go
tompizmor: actually, I don't need this right now. It's a leftover from a previous test using [oras library](https://github.com/deislabs/oras) to pull/push from/to OCI registries. I think we should be using that library, but I was getting out of time and wanted to have something working first.

pkg/client/oci/oci.go
tompizmor: they are valid responses for native OCI:

https://github.com/opencontainers/distribution-spec/blob/master/spec.md#endpoints

pkg/client/oci/oci.go
tompizmor: I just need to check for StatusOK, the rest of them should return an error. Do you think it still makes sense the switch?

api/config.pb.go
tompizmor: don't know how I had that devel version...🤷‍♂️

pkg/client/oci/oci.go
tompizmor: using this instead:

- https://github.com/opencontainers/image-spec/blob/master/specs-go/v1/manifest.go

pkg/client/oci/oci.go
```diff
@@ -0,0 +1,296 @@
...
+// getTagManifest returns the manifests of a published tag
+func (r *Repo) getTagManifest(name, version string) (*ocispec.Manifest, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), getTimeout)
+	defer cancel()
+
+	u := *r.url
+	// Form API endpoint URL from repo url
+	u.Path = path.Join("v2", u.Path, name, "manifests", version)
+	req, err := http.NewRequestWithContext(ctx, "GET", u.String(), nil)
+	req.Header.Set("Accept", "application/vnd.oci.image.manifest.v1+json")
```
jotadrilo, line 88: define it as a `const` and use it here

pkg/client/oci/oci.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+// getChartDigest returns the digest of a published chart
+func (r *Repo) getChartDigest(name, version string) (string, error) {
+	tm, err := r.getTagManifest(name, version)
+	if err != nil {
+		return "", errors.Trace(err)
+	}
+	for _, layer := range tm.Layers {
+		if layer.MediaType == "application/tar+gzip" {
```
jotadrilo, line 124:
```suggestion
		if layer.MediaType == HelmChartContentLayerMediaType {
```

pkg/client/oci/oci.go
```diff
@@ -0,0 +1,296 @@
...
+	status := resp.StatusCode
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, errors.Errorf("unexpected response — %d %q — from %s", status, http.StatusText(status), u.String())
+	}
+
+	switch status {
+	case http.StatusNotFound:
+		// If status not found 404, it could mean the asset has no release yet
+		return []string{}, nil
```
jotadrilo, line 164: please add a TODO here to tackle it later: Use the `NotFound` error instead of just returning `nil` and handle that case in the caller, instead.

pkg/client/oci/oci.go
```diff
@@ -0,0 +1,296 @@
...
+	if err := json.Unmarshal(body, ot); err != nil {
+		return nil, err
+	}
+	chartTags := []string{}
+	for _, tag := range ot.Tags {
+		tm, err := r.getTagManifest(name, tag)
+		if err != nil {
+			return nil, errors.Trace(err)
+		}
+		if tm.Config.MediaType == "application/vnd.cncf.helm.config.v1+json" {
```
jotadrilo, line 181:
```suggestion
		if tm.Config.MediaType == HelmChartConfigMediaType {
```

pkg/client/oci/oci.go
jotadrilo: please add a reference inline so others can check the list of endpoints

pkg/client/oci/oci.go
jotadrilo: yes, so it is simpler to add additional behaviors in the future and it is consistent with the other method

jotadrilo (approved):
LGTM!

PR #69, opened 2020-12-01
feat(internal): Mark some packages as internal

jotadrilo:
Currently, all packages are stored under the `pkg` folder. However, this folder is intended to contain packages that we might want to make accessible to other projects. 

This patch moves those that are not meant to be imported outside of this project to the `internal` folder.


tompizmor (approved):
lgtm

PR #71, opened 2020-12-02
feat(client): Extend Upload API to accept name and version fields

jotadrilo:
Both #67 and #70 require an extended version of `Upload` that accepts `name` and `version` chart values.


tompizmor (approved):
lgtm

PR #72, opened 2020-12-02
docs(readme): Update README after adding OCI support

tompizmor:
Just some minor updates on the README file after adding OCI support

### DIFF
--- a/charts-syncer.yaml
+++ b/charts-syncer.yaml
@@ -38,3 +38,9 @@ target:
       # password is the password used to authenticate against the target chart repo
       # `TARGET_AUTH_PASSWORD` env var can be used instead of this entry
       password: "PASSWORD"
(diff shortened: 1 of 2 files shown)


jotadrilo (approved):
LGTM!

charts-syncer.yaml
```diff
@@ -38,3 +38,9 @@ target:
       # password is the password used to authenticate against the target chart repo
       # `TARGET_AUTH_PASSWORD` env var can be used instead of this entry
       password: "PASSWORD"
+# charts is an OPTIONAL list to specify a subset of charts to be synchronized
+# It is mandatory iff the source repo is OCI as autodiscovery is not supported
```
jotadrilo, line 42: The "auto discovery" concept is something specific to the tool. I would mention that OCI does not support listing charts only.

PR #76, opened 2020-12-06
feat: Remove helm requirement

jotadrilo:
This patch allows packaging a chart without using the `helm` client.


pkg/syncer/sync.go
```diff
@@ -8,91 +8,15 @@ import (
 
 	"github.com/juju/errors"
 	"github.com/mkmik/multierror"
+	helm "helm.sh/helm/v3/pkg/action"
```
Javirln, line 11: Is this a leftover?

jotadrilo: no! Take into account that now we are doing:

```
		pkgCli := helm.NewPackage()
```

and I thought that `action.NewPackage` was a bit open for this context.

Javirln (approved):

PR #79, opened 2020-12-21
feat(client): Add local client impl

jotadrilo:
This feature will allow us to test the synchronization in local mode without a real implementation nor chart repo up and running.

```
❯ ./dist/charts-syncer sync --config examples/local.yaml
I1221 22:24:18.766701   61084 sync.go:33] Using config file: "examples/local.yaml"
I1221 22:24:37.459338   61084 sync.go:41] There are 653 charts out of sync!
I1221 22:24:37.459360   61084 sync.go:65] Syncing "redis-10.4.3" chart...
I1221 22:24:37.487419   61084 sync.go:65] Syncing "etcd-4.3.1" chart...
I1221 22:24:37.524282   61084 sync.go:65] Syncing "nginx-4.3.6" chart...
...
I1221 22:25:01.236403   61084 sync.go:65] Syncing "harbor-7.4.1" chart...
I1221 22:25:01.367822   61084 sync.go:65] Syncing "harbor-7.4.2" chart...
I1221 22:25:01.502106   61084 sync.go:65] Syncing "harbor-9.1.2" chart...

❯ tree local-repo
local-repo
├── common-0.10.0.tgz
├── common-0.3.1.tgz
...
├── redis-9.5.4.tgz
└── redis-9.5.5.tgz

0 directories, 653 files
```


migmartri (approved):
Not for this diff but while reading the Readme and playing around, I found quite confusing the different categories we support. I mean, Harbor itself can have OCI or chartmuseum as backend so I was not sure what to use.

migmartri:
What's the usecase for this? Test out the modifications of the charts?

jotadrilo:
> Not for this diff but while reading the Readme and playing around, I found quite confusing the different categories we support. I mean, Harbor itself can have OCI or chartmuseum as a backend so I was not sure what to use.

I agree. I would like to support some sort of auto-discovering mechanism, but I am afraid that's not possible. Then, we have this OCI standard that many backends may support (Harbor is one of those), while we could even (and we want to) support HARBOR_OCI because it offers more features (like listing one single kind of artifact at a time) in top of the regular OCI API implementation. What's your take on this? Would you clarify different use cases depending on the target? cc/ @tompizmor

jotadrilo:
> What's the usecase for this? Test out the modifications of the charts?

It is testing mainly yes. I thought of supporting this if and only if we provide a feature flag (like `ENABLE_TEST_FEATURES=1`) so we don't support this mechanism in a regular execution of the tool.

tompizmor (approved):
LGTM

tompizmor:
> I mean, Harbor itself can have OCI or chartmuseum as backend so I was not sure what to use.

This is basically because when we started this project Harbor only supported chartmuseum as backend. OCI support was added later.

Issue #73, opened 2020-12-02
Use oras library to push charts to OCI registry


tompizmor:
We are using helm CLI to push charts to the OCI registry. In order to remove helm as an external dependency, we should use [oras library](https://github.com/deislabs/oras) as it is what helm CLI uses internally.

PR #83, opened 2021-01-19
feat(helmcli): implement helmcli methods as an interface for ease testing

tompizmor:
By moving the helm CLI methods to an interface we can easily mockup those functions for unit testing using testify/mock package.

Additionally, I have renamed the package name from `helmcli` to `helm` and I have deleted unused methods


tompizmor:
Closing this PR in favor of #84. In the end, as collected on #73 our final goal was to not depend on helm cli.

PR #85, opened 2021-01-25
feat(tar): remove tar dependency

### LINKED ISSUE #45: Replace "tar" binary requirement by Go native libs
We are currently invoking the `tar` binary (so it is required that this binary is present in the machine) to extract the Chart content (see `pkg/utils`). We should replace it with a Go native implementation.

tompizmor:
This PR just re-implements the Untar function from the utils package to not depends on `tar` system command. Instead it makes used of golang native libraries.

Fixes #45

### DIFF
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -1,6 +1,8 @@
 package utils
 
 import (
+	"archive/tar"
+	"compress/gzip"
 	"crypto/sha1"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


jotadrilo (commented):
Please add a unit tests for the new functionality!

jotadrilo (commented):
LGTM!

jotadrilo (approved):

PR #86, opened 2021-01-25
test: improve Untar unit tests

tompizmor:
This PR just improves a bit the Untar unit tests so we also check that files from subfolders in the tarball are properly uncompressed.


jotadrilo (approved):
LGTM!

PR #84, opened 2021-01-20
feat(oras): Remove helm cli dependency

tompizmor:
This PR deletes the `helmcli` package and so the helm cli requirement. Previously we were only using it for pushing charts to  OCI repositories, from now on, we can use oras library directly.

Fixes https://github.com/bitnami-labs/charts-syncer/issues/73

PR #87, opened 2021-01-25
feat: use from scratch in Dockerfile

tompizmor:
After removing external helm cli and tar dependencies in #84 and #85 we can simply use a scratch container

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,11 @@
+FROM bitnami/minideb:buster as build
+RUN install_packages ca-certificates
+RUN mkdir /workdir
+
 FROM scratch
 ARG IMAGE_VERSION
 ENV IMAGE_VERSION=${IMAGE_VERSION}
+COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+COPY --from=build /workdir /tmp
 COPY ./charts-syncer /
-CMD [ "/charts-syncer" ]
+ENTRYPOINT [ "/charts-syncer" ]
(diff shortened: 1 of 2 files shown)


tompizmor:
Do not land, maybe this doesn't work after all 😢

tompizmor:
First I got an issue when downloading the index because of missing certs:

```
Error: reloading "https://charts.bitnami.com/bitnami" chart repo: fetching index.yaml: Get "https://charts.bitnami.com/bitnami/index.yaml": x509: certificate signed by unknown authority
```

I see that other scratch container uses a multistage build and copy the certs from the builder image. After doing that however, the next issue appeared:

```
Error: reloading "https://charts.bitnami.com/bitnami" chart repo: open /tmp/index.797316597.yaml: no such file or directory
```

tompizmor:
In the end, I was able to solve the non-existent `/tmp` folder by creating an empty folder in the build image and copying it to its final location as a multistage step.

Dockerfile
```diff
@@ -1,5 +1,11 @@
-FROM bitnami/minideb:buster
+FROM bitnami/minideb:buster as build
+RUN install_packages ca-certificates
+RUN mkdir /workdir
+
+FROM scratch
 ARG IMAGE_VERSION
 ENV IMAGE_VERSION=${IMAGE_VERSION}
+COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+COPY --from=build /workdir /tmp
```
jotadrilo, line 9: should we add a comment saying that this a workaround to create a `/tmp` folder in a scratch container?

jotadrilo (approved):

Dockerfile
tompizmor: added

PR #92, opened 2021-01-27
test(local): Added unit tests for local client

tompizmor:
This PR adds unit tests for the local client

### DIFF
--- a/pkg/client/local/local_test.go
+++ b/pkg/client/local/local_test.go
@@ -1,0 +1,117 @@
+package local
+
+import (
+	"os"
+	"reflect"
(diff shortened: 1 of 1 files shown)


pkg/client/local/local_test.go
```diff
@@ -0,0 +1,117 @@
+package local
```
jotadrilo, line 1: If you are not testing internal methods, please use `local_test` and import the `local` package here. I haven't tell you in other PRs, sorry.

jotadrilo (approved):
LGTM!

PR #94, opened 2021-01-28
feat(harbor): Add unit tests for harbor clients

tompizmor:
This PRs adds unit tests for the harbor client implementation

### DIFF
--- a/pkg/client/harbor/harbor.go
+++ b/pkg/client/harbor/harbor.go
@@ -27,7 +27,7 @@ type Repo struct {
 	username string
 	password string
 
-	Helm *helmclassic.Repo
+	helm *helmclassic.Repo
 
 	cache cache.Cacher
 }
@@ -49,7 +49,7 @@ func NewRaw(u *url.URL, user string, pass string, c cache.Cacher) (*Repo, error)
 		return nil, errors.Trace(err)
 	}
 
-	return &Repo{url: u, username: user, password: pass, Helm: helm, cache: c}, nil
+	return &Repo{url: u, username: user, password: pass, helm: helm, cache: c}, nil
 }
 
 // GetUploadURL returns the URL to upload a chart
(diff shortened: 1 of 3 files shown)


pkg/client/harbor/harbor.go
```diff
@@ -27,7 +27,7 @@ type Repo struct {
 	username string
 	password string
 
-	helm *helmclassic.Repo
+	Helm *helmclassic.Repo
```
jotadrilo, line 30: You should not expose this in purpose to ease your testing. The `harbor` client is exposing an interface and your tests (the ones under `harbor_test` package) should test the API.

If you need to test the internals, you can define internal tests (under `harbor` package) so you can access this kind of thing.

jotadrilo (changes requested):

pkg/client/harbor/harbor_test.go
```diff
@@ -0,0 +1,217 @@
...
+	harborRepo = &api.Repo{
+		Kind: api.Kind_HARBOR,
+		Auth: &api.Auth{
+			Username: "user",
+			Password: "password",
+		},
+	}
+)
+
+func prepareTest(t *testing.T) (*harbor.Repo, error) {
```
jotadrilo, line 35: you should consider using [testing.M](https://golang.org/pkg/testing/#hdr-Main). See [this article](https://www.ardanlabs.com/blog/2019/10/integration-testing-in-go-set-up-and-writing-tests.html#creating-testmain-using-testing-m) for a complete example.

jotadrilo (approved):

pkg/client/harbor/harbor_test.go
tompizmor: ok, I will check it after merging this PR and create a new one applying the changes to all the affected clients if it applies nicely

PR #95, opened 2021-01-28
feat(oci): Add unit tests for oci clients

tompizmor:
These PRs adds unit tests for OCI clients.

For these OCI tests, we need two different testing services. 

1. An HTTP server to handle direct HTTP requests made by the tool as an OCI server would do.
2. An OCI compliant server to test the push command made directly with oras library. We use docker-registry for that


pkg/client/oci/oci_test.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+// Starts an OCI compliant server (docker-registry) so our push command based on oras cli works out-of-the-box.
+// This way we don't have to mimic all the low-level HTTP requests made by oras.
+func prepareOciServer(t *testing.T) *oci.Repo {
+	t.Helper()
+
+	// Create OCI server as docker registry
+	config := &configuration.Configuration{}
+	port, err := freeport.GetFreePort()
```
jotadrilo, line 77: You don't need to use an external package for this. See the example below and feel free to add it to the `utils` package.

```
func getListenAddress(t *testing.T) string {
	t.Helper()

	lst, err := net.Listen("tcp", "127.0.0.1:0")
	if err != nil {
		t.Fatal(err)
	}
	defer lst.Close()

	return lst.Addr().String()
}
```

jotadrilo (approved):
LGTM!

go.mod
```diff
@@ -2,17 +2,23 @@ module github.com/bitnami-labs/charts-syncer
...
 	github.com/deislabs/oras v0.8.1
+	github.com/docker/distribution v2.7.1+incompatible
 	github.com/golang/protobuf v1.4.2
 	github.com/google/go-cmp v0.5.0
 	github.com/juju/errors v0.0.0-20200330140219-3fe23663418f
 	github.com/juju/testing v0.0.0-20200923013621-75df6121fbb0 // indirect
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/mkmik/multierror v0.3.0
 	github.com/opencontainers/image-spec v1.0.1
+	github.com/phayes/freeport v0.0.0-20180830031419-95f893ade6f2
```
jotadrilo, line 21: please run `go mod tidy`

pkg/client/oci/oci_test.go
```diff
@@ -0,0 +1,214 @@
...
+	addr, err := utils.GetListenAddress()
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, port, err := net.SplitHostPort(addr)
+	if err != nil {
+		t.Fatal(err)
+	}
+	dockerRegistryHost := "http://" + addr
+	config.HTTP.Addr = fmt.Sprintf(":%s", port)
```
jotadrilo, line 87: you can use the address from L78 directly. `:port` means "listen for port `port` in all interfaces"

tompizmor: rigth!

PR #97, opened 2021-01-29
fix(GetDownloadURL): Fix chart download url when url is relative in index.yaml

tompizmor:
While testing a replication from Harbor to Habor I notice this issue. We were returning the URL as it is in the index.yaml, however, this URL can be absolute or relative. 

**absolute**
```
 urls:
    - https://some.repo.url/charts/nginx-5.1.1.tgz
```

**relative**
```
 urls:
    - charts/nginx-5.1.1.tgz
```

We were assuming it was an absolute URL, so in the case it is not, we should transform it first:

### DIFF
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -272,3 +272,16 @@ func EncodeSha1(s string) string {
 	h.Write([]byte(s))
 	return fmt.Sprintf("%x", h.Sum(nil))
 }
+
+// FormDownloadURL forms the full download URL in case we pass a relative URL
+func FormDownloadURL(repoURL, indexURL string) (string, error) {
+	u, err := url.Parse(indexURL)
+	if err != nil {
+		return "", errors.Trace(err)
+	}
+	chartURL := u.String()
+	if u.Host == "" {
+		chartURL = fmt.Sprintf("%s/%s", repoURL, u.String())
+	}
+	return chartURL, nil
+}
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -273,15 +273,21 @@ func EncodeSha1(s string) string {
 	return fmt.Sprintf("%x", h.Sum(nil))
 }
 
-// FormDownloadURL forms the full download URL in case we pass a relative URL
-func FormDownloadURL(repoURL, indexURL string) (string, error) {
-	u, err := url.Parse(indexURL)
+// NormalizeChartURL forms the full download URL in case we pass a relative URL
+func NormalizeChartURL(repoURL, indexURL string) (string, error) {
+	iu, err := url.Parse(indexURL)
 	if err != nil {
 		return "", errors.Trace(err)
 	}
-	chartURL := u.String()
-	if u.Host == "" {
-		chartURL = fmt.Sprintf("%s/%s", repoURL, u.String())
+	ru, err := url.Parse(repoURL)
+	if err != nil {
+		return "", errors.Trace(err)
+	}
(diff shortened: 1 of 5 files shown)


pkg/client/helmclassic/helmclassic.go
```diff
@@ -109,7 +109,15 @@ func (r *Repo) GetDownloadURL(n string, v string) (string, error) {
...
-	return chart.URLs[0], nil
+	u, err := url.Parse(chart.URLs[0])
+	if err != nil {
+		return "", errors.Trace(err)
+	}
+	chartURL := u.String()
+	if u.Host == "" {
+		chartURL = fmt.Sprintf("%s/%s", r.url, u.String())
+	}
+	return chartURL, nil
```
jotadrilo, line 120: this chunk of logic deserves its own function, so you can add unit tests to it!

jotadrilo (changes requested):

pkg/client/helmclassic/helmclassic.go
tompizmor: just created a function in the utils package with specific unit tests

internal/utils/utils.go
```diff
@@ -272,3 +272,16 @@ func EncodeSha1(s string) string {
...
+// FormDownloadURL forms the full download URL in case we pass a relative URL
+func FormDownloadURL(repoURL, indexURL string) (string, error) {
+	u, err := url.Parse(indexURL)
+	if err != nil {
+		return "", errors.Trace(err)
+	}
+	chartURL := u.String()
+	if u.Host == "" {
+		chartURL = fmt.Sprintf("%s/%s", repoURL, u.String())
+	}
```
jotadrilo, line 285: You should also check that the host URL matches the index URL and return an error otherwise. I think this package is too generic. Please rename the function to `NormalizeChartURL`

jotadrilo (changes requested):

internal/utils/utils.go
tompizmor: done

internal/utils/utils_test.go
```diff
@@ -158,30 +159,48 @@ func TestIsValidURL(t *testing.T) {
...
+		expectedError error
 	}{
 		{
-			"full url index",
-			"https://chart.repo.url/charts/nats-1.2.3.tgz",
+			desc:          "full url index",
+			repoURL:       "https://chart.repo.url",
+			chartURL:      "https://chart.repo.url/charts/nats-1.2.3.tgz",
+			shouldFail:    false,
+			expectedError: nil,
```
jotadrilo, line 176: you can remove defaults

internal/utils/utils_test.go
```diff
@@ -158,30 +159,48 @@ func TestIsValidURL(t *testing.T) {
...
+			chartURL:      "charts/nats-1.2.3.tgz",
+			shouldFail:    false,
+			expectedError: nil,
+		},
+		{
+			desc:          "different hosts",
+			repoURL:       "https://chart.another-repo.url",
+			chartURL:      "https://chart.repo.url/charts/nats-1.2.3.tgz",
+			shouldFail:    true,
+			expectedError: errors.New("index URL host (chart.repo.url) and repo URL host (chart.another-repo.url) host are different"),
```
jotadrilo, line 190:
```suggestion
			expectedError: errors.New("index host (chart.repo.url) and repo host (chart.another-repo.url) are different"),
```

jotadrilo (approved):
LGTM!

PR #99, opened 2021-03-29
feat: add insecure flag

tompizmor:
This PR adds a `--insecure` command-line flag to allow skipping TLS checks for example when using self-signed certificate during development or testing.

Fixes https://github.com/bitnami-labs/charts-syncer/issues/78

### DIFF
--- a/cmd/sync.go
+++ b/cmd/sync.go
@@ -15,7 +15,6 @@ import (
 var (
 	syncFromDate string
 	syncWorkdir  string
-	insecure     bool
 )
 
 var (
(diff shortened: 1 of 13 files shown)


cmd/sync.go
```diff
@@ -87,7 +89,8 @@ func newSyncCmd() *cobra.Command {
 	}
 
 	cmd.Flags().StringVar(&syncFromDate, "from-date", "", "Date you want to synchronize charts from. Format: YYYY-MM-DD")
-	cmd.Flags().StringVar(&syncWorkdir, "workdir", syncer.DefaultWorkdir(), "Working directory.")
+	cmd.Flags().StringVar(&syncWorkdir, "workdir", syncer.DefaultWorkdir(), "Working directory")
+	cmd.Flags().BoolVarP(&insecure, "insecure", "", false, "Allow insecure SSL connections")
```
fjagugar, line 93: If we are not going to use the "shorthand letter" param, why not using "BoolVar" instead of "BoolVarP" (https://github.com/spf13/pflag/blob/master/bool.go#L53)

fjagugar (approved):

PR #102, opened 2021-08-17
Add the ability to use with github.io and github.com charts

marthydavid:
Today i've started to use this solution and I have encountered a problem with it.

I've tried to sync ingress-nginx helm charts but the following showed up.

```bash
kubectl logs -f charts-syncer-test-ltstj
I0817 11:05:34.533952       1 sync.go:33] Using config file: "/charts-syncer.yaml"
I0817 11:05:34.534501       1 syncer.go:96] Using workdir: "/.charts-syncer"
I0817 11:05:34.534744       1 cache.go:38] Allocating cache dir: "/.charts-syncer/077ebcf1faa6f9081bce292dd5a7de089a21971d"
I0817 11:05:34.534877       1 helmclassic.go:48] [0de643e32b4220f64888abca0464621f42fe743e] GET "https://kubernetes.github.io/ingress-nginx/index.yaml"
I0817 11:05:39.651954       1 helmclassic.go:66] [0de643e32b4220f64888abca0464621f42fe743e] HTTP Status: 200 OK
I0817 11:05:39.685720       1 cache.go:38] Allocating cache dir: "/.charts-syncer/5b6fde763f9acb2fe436de40e9ffbc348b2e7e0c"
I0817 11:05:39.685913       1 helmclassic.go:48] [c307586fa56b69b76502f3812cb493a79f952011] GET "https://harbor.my.tld/chartrepo/internal-repo/index.yaml"
I0817 11:05:40.046051       1 helmclassic.go:66] [c307586fa56b69b76502f3812cb493a79f952011] HTTP Status: 200 OK
I0817 11:05:40.280086       1 index.go:85] Publishing threshold set to "1970-01-01 00:00:00 +0000 UTC"
I0817 11:05:40.280128       1 index.go:97] Indexing "ingress-nginx" charts...
I0817 11:05:40.280203       1 cache.go:54] cache hit { op:has, id:077ebcf1faa6f9081bce292dd5a7de089a21971d, filename:ingress-nginx-4.0.0-beta.3.tgz }
E0817 11:05:40.280274       1 index.go:127] unable to load "ingress-nginx-4.0.0-beta.3" chart: index host (github.com) and repo host (kubernetes.github.io) are different
I0817 11:05:40.280352       1 cache.go:54] cache hit { op:has, id:077ebcf1faa6f9081bce292dd5a7de089a21971d, filename:ingress-nginx-4.0.0-beta.2.tgz }
[...]

PR #103, opened 2021-09-02
fix: Fix how we normalize chart URLs

jotadrilo:
We have been reported at #102 that the current logic does not support providing an index that refers to charts using a different host to the one hosting the index YAML file.

We didn't consider this use case due to a bug in the `NormalizeChartURL` method at the `internal/utils` package. After checking the PR, we noticed that our CI tests were not being triggered and we started looking into the issue.

This patch fixes the bug and other omitted edge cases and we can consider #102 as closed after merging it.

Nevertheless, we would like to appreciate the effort, time, and patience that @marthydavid has dedicated! :)


marthydavid:
Thanks!

Tested this morning at it works.

PR #105, opened 2021-09-17
fix(package-vulnerability) Update helm/v3 and containerd package version

tompizmor:
Dependabot notifiy about some vulnerabilities in these packages so let's update them

Screenshot 2021-09-17 at 15 32 34


jotadrilo (approved):
LGTM!

PR #106, opened 2021-09-17
chore: update circleci config

tompizmor:
We were using a wrong version of go and also installing `helm` cli when it was not needed at that point


jotadrilo:
LGTM!

PR #111, opened 2021-11-24
feature: add support for OCI index

tompizmor:
Allow using OCI index file to sync from OCI charts repositories. This feature brings the following advantages:

1. No need to specify the list of charts to synchronise in config.yaml file.
2. The process is now much faster as we don't need to autodiscover which tags are charts or other kind of assets. Previously  charts-syncer had to download the OCI manifest for every artifact tag to check of the mimeType corresponds to a chart.

More info about the usage is documented in the README.


api/oci.proto
```diff
@@ -0,0 +1,45 @@
+syntax = "proto3";
+
+package api;
+option go_package = "github.com/bitnami-labs/charts-syncer/api;api";
+
+
+message OciIndex {
```
migmartri, line 7: I'd add a comment here explaining that this is a proposed custom index

README.md
```diff
@@ -150,7 +150,48 @@ target:
...
+In order to speed up the sync process for OCI charts, it is possible to specify a reference to an OCI index file containing 
+the chart versions to be synced from the repository. Thanks to the index, chart-syncer don't need to autodiscover 
+what tags for a specific asset are charts or other kind of artifacts which is a costly process.
+
+~~~yaml
+source:
+ repo:
+   kind: OCI
+   url: https://my-oci-registry.io/my-project/subpath
+   ociIndex: my-oci-registry.io/my-project/index:latest
```
migmartri, line 167: Maybe ociIndex is not the best name because it does not explain what's indexing.

Technically it's a `chartsIndex`

README.md
```diff
@@ -150,7 +150,48 @@ target:
 `subpath` in the previous url is optional in case your charts are not stored directly under your projects.
 It is worth mentioning that you can use Harbor robot accounts using OCI registries as source or target.
 
-Also, take into account that if you use OCI as the source repository you must specify the list of chart to synchronize
+Also, take into account that if you use OCI as the source repository you must specify the list of charts to synchronize
+or a pointer to an [index file](#oci-index) in the repository.
+
+#### OCI index
```
migmartri, line 156: I'd be more verbose in the wording here, something like  "Helm Charts Index for OCI-based repositories"

README.md
```diff
@@ -150,7 +150,48 @@ target:
...
+
+In order to speed up the sync process for OCI charts, it is possible to specify a reference to an OCI index file containing
```
migmartri, line 158: To me, the main value of this feature is not speeding up the process but auto-discover what charts to synchronize without maintaining a hardcoded list.

The speed side-effect is a nice one but it's just an optimization.

README.md
```diff
@@ -150,7 +150,48 @@ target:
...
 
-Also, take into account that if you use OCI as the source repository you must specify the list of chart to synchronize
+Also, take into account that if you use OCI as the source repository you must specify the list of charts to synchronize
+or a pointer to an [index file](#oci-index) in the repository.
+
+#### OCI index
+
+In order to speed up the sync process for OCI charts, it is possible to specify a reference to an OCI index file containing 
+the chart versions to be synced from the repository. Thanks to the index, chart-syncer don't need to autodiscover 
+what tags for a specific asset are charts or other kind of artifacts which is a costly process.
```
migmartri, line 160: Yeah, I'd not mention this in the Readme, you are explaining implementation details.

In this section I'd just say that you make chart-syncer auto-discover what Helm Charts to sync by providing the URI of a custom index in the OCI registry.

README.md
```diff
@@ -150,7 +150,48 @@ target:
...
+~~~yaml
+source:
+ repo:
+   kind: OCI
+   url: https://my-oci-registry.io/my-project/subpath
+   ociIndex: my-oci-registry.io/my-project/index:latest
+~~~
+
+You can specify the index in the format of `REGISTRY/PROJECT/[SUBPATH][:TAG|@sha256:DIGEST]`. If the `ociIndex` property is provided,
+charts-syncer will use it directly, otherwise it will try to guess it from the URL field by looking for an index:latest asset located
```
migmartri, line 171: > otherwise it will try to guess it from the URL field by looking for an index:latest

This sentence imo could be renamed to but "otherwise it will fallback to retrieve the index from a default location `index:latest`" since it does not guess it right?

pkg/client/oci/oci.go
```diff
@@ -62,18 +72,75 @@ func New(repo *api.Repo, c cache.Cacher, insecure bool) (*Repo, error) {
...
+	resolver := newDockerResolver(u, repo.GetAuth().GetUsername(), repo.GetAuth().GetPassword(), insecure)
+	// Init entries
+	entries := make(map[string][]string)
 
-	return NewRaw(u, repo.GetAuth().GetUsername(), repo.GetAuth().GetPassword(), c, insecure)
+	// First we try to load the index referenced from config file,
+	// if it does not exist, then we try to guess it directly from repo url
+	ociIndexRef := repo.GetOciIndex()
+	if ociIndexRef == "" {
+		klog.Infof("OCI index reference is empty in config file. Guessing index from repo url")
```
migmartri, line 83: I mentioned above but I do not think `guessing` is the right word but instead  something like

No Charts OCI index reference specified. Checking default location %s"

pkg/client/oci/oci.go
```diff
@@ -62,18 +72,75 @@ func New(repo *api.Repo, c cache.Cacher, insecure bool) (*Repo, error) {
...
+	}
+
+	ociIndexExists, err := ociReferenceExists(u, ociIndexRef, repo.GetAuth().GetUsername(), repo.GetAuth().GetPassword(), insecure)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+
+	if ociIndexExists {
+		// Pull with ORAS lib
+		filename := "asset-index.json"
```
migmartri, line 94: Could we encapsulate this in a function?

pkg/client/oci/oci.go
```diff
@@ -62,18 +72,75 @@ func New(repo *api.Repo, c cache.Cacher, insecure bool) (*Repo, error) {
...
 
 // NewRaw creates a Repo object.
-func NewRaw(u *url.URL, user string, pass string, c cache.Cacher, insecure bool) (*Repo, error) {
-	return &Repo{url: u, username: user, password: pass, cache: c, insecure: insecure}, nil
+func NewRaw(u *url.URL, user string, pass string, c cache.Cacher, insecure bool, entries map[string][]string, resolver remotes.Resolver) (*Repo, error) {
+	return &Repo{url: u, username: user, password: pass, cache: c, insecure: insecure, entries: entries, dockerResolver: resolver}, nil
 }
 
 // List lists all chart names in a repo
 func (r *Repo) List() ([]string, error) {
```
migmartri, line 133: nice :partyparrot

pkg/client/oci/oci.go
```diff
@@ -138,6 +205,10 @@ func (r *Repo) getChartDigest(name, version string) (string, error) {
 
 // ListChartVersions lists all versions of a chart
 func (r *Repo) ListChartVersions(name string) ([]string, error) {
+	// If entries is populated use it to list the chart versions
```
migmartri, line 208: I'd extend this comment with some of the info you put in the readme, what you explained that if you don't use the index listing versions require checking the tags and so on...

pkg/client/oci/oci.go
```diff
@@ -354,9 +425,85 @@ func (r *Repo) Reload() error {
...
-func (r *Repo) newDockerResolver() remotes.Resolver {
+
+// OciReference contains info about an oci reference
+type ociReference struct {
+	path string
+	tag string
+}
+
+// splitOciReference split the tag or digest from an oci reference if exists
+func splitOciReference(ociRef string) ociReference {
```
migmartri, line 436: we probably don't want to add another dependency but in relok8s we use this library to manipulate OCI references, in case it's useful https://pkg.go.dev/github.com/google/go-containerregistry@v0.7.0/pkg/name#ParseReference

pkg/client/oci/oci.go
```diff
@@ -354,9 +425,85 @@ func (r *Repo) Reload() error {
...
+		return ociReference{path: res[1], tag: res[2]}
+	}
+
+	return ociReference{path: ociRef}
+}
+
+// ociReferenceExists checks if a given oci reference exists in the repository
+// OCI endpoints: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#endpoints
+func ociReferenceExists(repoUrl *url.URL, ociRef, username, password string, insecure bool) (bool, error) {
+	ociReference := splitOciReference(ociRef)
```
migmartri, line 455: by using go-container-registry it seems you'll get access to some of these check methods all of them using image.reference as receiver

https://pkg.go.dev/github.com/google/go-containerregistry@v0.7.0/pkg/v1/remote#Head

migmartri (commented):
This is awesome @tompizmor!

Most of my comments have been around how to explain the feature in the Readme. I'd also check if we could utilize https://pkg.go.dev/github.com/google/go-containerregistry to simplify the parsing and checking code.

api/config.proto
```diff
@@ -30,6 +30,8 @@ message Repo {
     Auth auth = 3;
     // The path where the repo stores charts. Useful for LOCAL kind only
     string path = 4;
+    // The uri where the OCI repository keeps the index of charts
+    string ociIndex = 5;
```
jotadrilo, line 34: I would rename this to `chartsIndex` because OCI is just one implementation where we are allowing users to store this in-house index. We could potentially add/retrieve this index to helm classic registries if we want.

api/oci.proto
jotadrilo: +1

pkg/client/core/core_test.go
```diff
@@ -2,12 +2,22 @@ package core
 
 import (
 	"fmt"
+	"github.com/bitnami-labs/charts-syncer/pkg/client/oci"
```
jotadrilo, line 5: run `goimports`

pkg/client/oci/oci.go
```diff
@@ -62,18 +72,75 @@ func New(repo *api.Repo, c cache.Cacher, insecure bool) (*Repo, error) {
...
+	// Init entries
+	entries := make(map[string][]string)
 
-	return NewRaw(u, repo.GetAuth().GetUsername(), repo.GetAuth().GetPassword(), c, insecure)
+	// First we try to load the index referenced from config file,
+	// if it does not exist, then we try to guess it directly from repo url
+	ociIndexRef := repo.GetOciIndex()
+	if ociIndexRef == "" {
+		klog.Infof("OCI index reference is empty in config file. Guessing index from repo url")
+		ociIndexRef = fmt.Sprintf("%s%s/index:latest", u.Host, u.Path)
```
jotadrilo, line 84: as follow up in the rename suggestion I made, I'd say:

```suggestion
		chartsIndexRef = fmt.Sprintf("%s%s/index:latest", u.Host, u.Path)
		klog.Infof("Charts index reference has not been set in the config file. Checking default location: %q", chartsIndexRef)
```

pkg/client/oci/oci.go
```diff
@@ -62,18 +72,75 @@ func New(repo *api.Repo, c cache.Cacher, insecure bool) (*Repo, error) {
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
+	resolver := newDockerResolver(u, repo.GetAuth().GetUsername(), repo.GetAuth().GetPassword(), insecure)
+	// Init entries
+	entries := make(map[string][]string)
```
jotadrilo, line 77: Initialize these variables closer to the context where they are used for the first time

pkg/client/oci/oci.go
```diff
@@ -62,18 +72,75 @@ func New(repo *api.Repo, c cache.Cacher, insecure bool) (*Repo, error) {
...
 
 // List lists all chart names in a repo
 func (r *Repo) List() ([]string, error) {
-	return nil, errors.Errorf("list method is not supported yet")
+	var names []string
+	// If entries is not populated, it means we couldn't load any index file so we need the charts filter in the
+	// configuration file
+	if len(r.entries) == 0 {
+		return nil, errors.New("No OCI index file detected. Please provide the charts filter in config file")
+	}
```
jotadrilo, line 139: I think this is out of the scope of this method's purpose. It is OK to return no items in a `#List` method. The caller must handle this case and perform an action accordingly. Actually, the OCI index is not a cause either since an index might be empty due to having no charts in the registry.

jotadrilo (changes requested):
I have added a few comments so far, I will continue later

pkg/client/oci/oci.go
jotadrilo: if there is existing libs for that, I'd use them

pkg/client/oci/oci.go
jotadrilo: +1 if there is a lib that manages it

jotadrilo (changes requested):
Overall, please configure your IDE/editor to format files on save and re-organize imports to have:

```
import (
    "native/imports"

    "external.domain/imports"
)

File: .circleci/config.yml [hunk 18]
```diff
@@ -46,7 +46,7 @@ workflows:
 jobs:
   build:
     docker:
-      - image: circleci/golang:1.15
+      - image: circleci/golang:1.16
```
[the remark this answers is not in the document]
tompizmor: Not yet! I tried but there are some weird errors in the tests. I would do it later on a separate PR

api/config.proto
tompizmor: Done! The same was suggested by @migmartri

pkg/client/oci/oci.go
tompizmor: Added some changes related to this. 

We were only calling to oci.List() method from index.go (loadCharts() method) if charts filter was not provided in the config file. So once we reach the oci.List() method, if entries is empty means we couldn't load any index file and so we have no idea what to sync. That's why I was returning the error at oci.List() method before

Now, oci.List() returns an empty list if entries is empty and the caller (loadCharts() methods) will check this specific case (source is OCI, no entries and no charts filter) and will throw the error. Note I added an abortExecution flag to finish charts-syncer execution depending on the error reported in loadCharts(). For some of them we want the execution to continue as they are just errors for an specific chart version.

README.md
```diff
@@ -150,7 +150,47 @@ target:
 `subpath` in the previous url is optional in case your charts are not stored directly under your projects.
 It is worth mentioning that you can use Harbor robot accounts using OCI registries as source or target.
 
-Also, take into account that if you use OCI as the source repository you must specify the list of chart to synchronize
+Also, take into account that if you use OCI as the source repository you must specify the list of charts to synchronize
+or a pointer to an [index file](#helm-charts-index-for-oci-based-repositories) in the repository.
+
+#### Helm Charts Index for OCI-based repositories
+
+By using a charts index file for OCI-bases repository you won't need to maintain a hardcoded list of chart names in the config file.
```
migmartri, line 158: OCI-Based

README.md
```diff
@@ -150,7 +150,47 @@ target:
...
+charts-syncer will be able to auto-discover what charts need to be synchronized.
```
migmartri, line 159: out of curiosity, if I put the filter in the file but the index is also there what happens? Does it work as a filter?

README.md
```diff
@@ -150,7 +150,47 @@ target:
...
+~~~
+
+You can specify the index in the format of `REGISTRY/PROJECT/[SUBPATH][:TAG|@sha256:DIGEST]`. If the `chartsIndex` property is provided,
+charts-syncer will use it directly, otherwise it will fallback to retrieve the index from a default location `index:latest` asset located
+directly at URL level.
+
+For example, if your URL is `https://my-oci-registry.io/my-project/subpath` and no `chartsIndex` is specified, charts-syncer will try to use 
+`my-oci-registry.io/my-project/subpath/index:latest` asset as index if it exists.
+
+An example of the valid index format can be seen directly in its [protobuffer definition](api/oci.proto).
```
migmartri, line 176: I'd also mention here that the index format is custom, not a traditional Helm index file.

README.md
```diff
@@ -150,7 +150,47 @@ target:
...
+You can specify the index in the format of `REGISTRY/PROJECT/[SUBPATH][:TAG|@sha256:DIGEST]`. If the `chartsIndex` property is provided,
+charts-syncer will use it directly, otherwise it will fallback to retrieve the index from a default location `index:latest` asset located
+directly at URL level.
+
+For example, if your URL is `https://my-oci-registry.io/my-project/subpath` and no `chartsIndex` is specified, charts-syncer will try to use 
+`my-oci-registry.io/my-project/subpath/index:latest` asset as index if it exists.
+
+An example of the valid index format can be seen directly in its [protobuffer definition](api/oci.proto). 
+
+Finally, if no OCI index is found, charts-syncer would still need the list of charts in the config file:
```
migmartri, line 178: will require

api/oci.proto
```diff
@@ -0,0 +1,48 @@
...
+//      "versions": [
+//        {
+//          "version": "2.2.2",
+//          "digest": "sha256:jki0ea2a722b455851ff4688ea5bed2bac1992b7c72aece178bd12e3a70cetra",
+//          "uri": "my-oci-registry.io/my-project/charts/asset-2:2.2.2"
+//        }
+//      ]
+//    }
+//  ]
+//}
```
migmartri, line 48: new line missing here

pkg/client/oci/oci.go
```diff
@@ -40,14 +46,20 @@ const (
 	ImageManifestMediaType = "application/vnd.oci.image.manifest.v1+json"
 )
 
+var (
+	tagRegex = regexp.MustCompile(`^([a-z0-9.:]+/[\w\-_/]+):?([\w0-9.]+)?`)
```
migmartri, line 50: is this being used?

pkg/client/oci/oci.go
```diff
@@ -354,9 +405,113 @@ func (r *Repo) Reload() error {
...
 }
 
-func (r *Repo) newDockerResolver() remotes.Resolver {
+// SetEntries sets the entries of a repo
+func (r *Repo) SetEntries(entries map[string][]string) {
+	r.entries = entries
+}
+
+// OciReference contains info about an oci reference
+type ociReference struct {
```
migmartri, line 414: being used?

pkg/client/oci/oci.go
```diff
@@ -354,9 +405,113 @@ func (r *Repo) Reload() error {
 	return errors.Errorf("reload method is not supported yet")
 }
 
-func (r *Repo) newDockerResolver() remotes.Resolver {
+// SetEntries sets the entries of a repo
+func (r *Repo) SetEntries(entries map[string][]string) {
```
migmartri, line 409: It seems that you are creating this method just to enable testing from a different package `oci_test`. I personally don't like that. it usually means that we are mixing up unit testing with interface/functional testing. It also pollutes the package public interface with a method that's not meant to be used outside of testing.

Is there any other way to achieve the same effect without relying on this?

pkg/client/oci/oci.go
```diff
@@ -354,9 +405,113 @@ func (r *Repo) Reload() error {
...
+		ociReferenceIdentifier = parts[1]
+	}
+	contextOciRef := fmt.Sprintf("%s://%s", repoUrl.Scheme, ociReference.Context())
+	u, err := url.Parse(contextOciRef)
+	if err != nil {
+		return false, errors.Trace(err)
+	}
+	u.Path = path.Join("v2", u.Path, "manifests", ociReferenceIdentifier)
+
+	req, err := http.NewRequest("HEAD", u.String(), nil)
```
migmartri, line 438: as I mentioned in a previous comment. Have you tried to use https://pkg.go.dev/github.com/google/go-containerregistry@v0.7.0/pkg/v1/remote#Head ?

That one seems to require a `name.Referece` which you already have from line 422.

pkg/syncer/index.go
```diff
@@ -62,15 +63,20 @@ func (i ChartIndex) Get(id string) *Chart {
 }
 
 // loadCharts loads the charts map into the index from the source repo
-func (s *Syncer) loadCharts(charts ...string) error {
+func (s *Syncer) loadCharts(charts ...string) (bool, error) {
 	if len(charts) == 0 {
 		if !s.autoDiscovery {
-			return errors.Errorf("unable to discover charts to sync")
+			return true, errors.Errorf("unable to discover charts to sync")
```
migmartri, line 69: can you add a comment mentioning than the boolean means abort execution?

pkg/syncer/index.go
```diff
@@ -62,15 +63,20 @@ func (i ChartIndex) Get(id string) *Chart {
...
 		}
-
 		srcCharts, err := s.cli.src.List()
+		// For OCI source we need either access to a charts index file in the repo or a list of charts provided via
+		// config file
+		if len(srcCharts) == 0 && s.source.GetRepo().GetKind() == api.Kind_OCI {
```
migmartri, line 74: I don't understand what has changed that makes it different than before? 

Is it so we can detect if the index was empty?

pkg/syncer/sync.go
```diff
@@ -26,7 +26,11 @@ func (s *Syncer) SyncPendingCharts(names ...string) error {
 	// There might be problems loading all the charts due to missing dependencies,
 	// invalid/wrong charts in the repository, etc. Therefore, let's warn about
 	// them instead of blocking the whole sync.
-	if err := s.loadCharts(names...); err != nil {
+	// If there is a severe error, abortExecution would be true, and we would end the whole sync (with proper error code)
+	abortExecution, err := s.loadCharts(names...)
```
migmartri, line 30: do we really need to abort? It seems that the only case now is when the index is empty? If so is that smth we need to check?

migmartri (approved):

migmartri (approved):

pkg/client/oci/oci.go
tompizmor: I just tried it but there is something I don't like:

Currently I manage three situations based on exitCode:

- index exists. code 200
- index not exists. code 404
- other status are treated as error. For example wrong credentials

with remote.Head() method if the index does not exists directly throws an error:

```
Error: HEAD https://my-oci-registry.io/v2/my-project/charts/index/manifests/latest: unexpected status code 404 Not Found (HEAD responses have no body, use GET for details)
```

However, I don't want to throw an error if the index does not exist

pkg/syncer/index.go
tompizmor: Two things:
- the oci.List() method can return an empty list of charts as per jdrios comment
- we can now detect which errors coming from loadCharts function should abort the execution and which don't. Previously all errors were reported at the end but charts-syncer ended with exit code 0.

pkg/syncer/sync.go
tompizmor: I am trying to protect the case where no index file is provided or found in the repo and no charts filter is specified. If that is the case, charts-syncer would just say there is nothing to sync. I would like tool to provide feedback about a possible misconfiguration although we can just leave it as it was.

migmartri: I see and I agree with making sure that the user provides one or the other. What I am not sure though is if the code responsible of that guard clause should be here. It looks to me (although I have very limited knowledge on this codebase) that using this boolean as guard clause could be confusing.

Technically it could be used for other things i.e connectivity is gone and we do not want to retry.

I am wondering if there is a more explicit way up in the stack that we could make that check. Couldn't we just check if the index exists || the filter is set in the oci.New function (where we already check if the index exists?)

migmartri (approved):
Added some minor comments but LGTM!

pkg/client/oci/oci.go
```diff
@@ -354,9 +405,113 @@ func (r *Repo) Reload() error {
...
+	// Read downloaded file
+	ociIndexBytes, err := ioutil.ReadFile(filename)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	r := bytes.NewReader(ociIndexBytes)
+
+	// Unmarshall index and populate entries
+	ociIndex := &api.OciIndex{}
+	err = pbjson.NewDecoder(r).Decode(ociIndex)
```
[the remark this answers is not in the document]
tompizmor: Done

pkg/client/oci/oci.go
```diff
@@ -354,9 +405,113 @@ func (r *Repo) Reload() error {
...
+		bodyStr := utils.HTTPResponseBody(res)
+		return false, errors.Errorf("unable to check if OCI index exists, got HTTP Status: %s, Resp: %v", res.Status, bodyStr)
+	}
+
+	return true, nil
+}
+
+// populateEntries populates the entries map with the info from the charts index
+func populateEntries(chartsIndexRef string, resolver remotes.Resolver) (map[string][]string, error) {
+	entries := make(map[string][]string)
```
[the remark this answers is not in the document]
tompizmor: Added!

pkg/client/oci/oci.go
```diff
@@ -354,9 +405,113 @@ func (r *Repo) Reload() error {
...
+	// Pull with ORAS lib
+	filename := "asset-index.json"
+	defer os.Remove(filename)
+	ctx := context.Background()
```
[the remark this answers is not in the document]
tompizmor: I updated the context to use a timeout

pkg/client/oci/oci_internal_test.go
```diff
@@ -0,0 +1,85 @@
...
+				t.Fatal(err)
+			}
+			if got != tc.want {
+				t.Errorf("wrong result from OCI reference existence check. got: %v, want: %v", got, tc.want)
+			}
+		})
+	}
+}
+
+func TestListWithEntries(t *testing.T) {
```
migmartri, line 66: Awesome :)

pkg/client/oci/oci.go
```diff
@@ -354,9 +402,67 @@ func (r *Repo) Reload() error {
...
+func ociReferenceExists(ociRef, username, password string) (bool, error) {
+	ociReference, err := name.ParseReference(ociRef)
+	if err != nil {
+		return false, errors.Trace(err)
+	}
+	authOptions := authn.Basic{Username: username, Password: password}
+	opt := []remote.Option{
+		remote.WithAuth(&authOptions),
+	}
+	_, err = remote.Head(ociReference, opt...)
```
migmartri, line 415: much simpler!

pkg/syncer/index.go
```diff
@@ -62,16 +60,20 @@ func (i ChartIndex) Get(id string) *Chart {
 }
 
 // loadCharts loads the charts map into the index from the source repo
+// The returned boolean means that we should abort the execution rather that keeping the error to show all of them
```
migmartri, line 63: this comment entry seems outdated

migmartri (approved):
LGTM!

Thanks for applying all that feedback!

pkg/syncer/index.go
tompizmor: good catch! deleted!

PR #112, opened 2021-11-24
feature: add version command

tompizmor:
This PR adds a new `version` command to output the version of the tool.

Additionally it updates the required go version to 1.16

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -8,6 +8,8 @@ builds:
       - GO111MODULE=on
     main: ./main.go
     binary: charts-syncer
+    ldflags:
+      - -X github.com/bitnami-labs/charts-syncer/cmd.Version={{.Version}}
 dockers:
   -
     ids:
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -9,7 +9,7 @@ builds:
     main: ./main.go
     binary: charts-syncer
     ldflags:
-      - -X github.com/bitnami-labs/charts-syncer/cmd.Version={{.Version}}
+      - -X github.com/bitnami-labs/charts-syncer/cmd.version={{.Version}}
 dockers:
   -
     ids:
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -1,0 +1,26 @@
+package cmd
+
+import (
+	"fmt"
+	"github.com/spf13/cobra"
+)
+
+var Version = "dev"
+
+func versionHelp() string {
+	return fmt.Sprint("Print the version number of charts-syncer")
+}
+
+func newVersionCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "version",
+		Short:  versionHelp(),
+		Long: versionHelp(),
+		Example: versionHelp(),
+		Run: func(cmd *cobra.Command, args []string) {
+			cmd.Printf("charts-syncer version: %s\n", Version)
+		},
+	}
+
+	return cmd
+}
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -1,24 +1,21 @@
 package cmd
 
 import (
-	"fmt"
 	"github.com/spf13/cobra"
 )
 
-var Version = "dev"
+var version = "dev"
 
 func versionHelp() string {
-	return fmt.Sprint("Print the version number of charts-syncer")
+	return "Print the version number of charts-syncer"
 }
 
 func newVersionCmd() *cobra.Command {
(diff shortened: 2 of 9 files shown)


.goreleaser.yml
```diff
@@ -8,6 +8,8 @@ builds:
       - GO111MODULE=on
     main: ./main.go
     binary: charts-syncer
+    ldflags:
+      - -X github.com/bitnami-labs/charts-syncer/cmd.Version={{.Version}}
```
tompizmor, line 12: goreleaser will take the version from the git tag

cmd/version.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"fmt"
+	"github.com/spf13/cobra"
+)
+
+var Version = "dev"
+
+func versionHelp() string {
+	return fmt.Sprint("Print the version number of charts-syncer")
```
jotadrilo, line 11:
```suggestion
	return "Print the version number of charts-syncer"
```

cmd/version.go
```diff
@@ -0,0 +1,26 @@
...
+
+func newVersionCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "version",
+		Short:  versionHelp(),
+		Long: versionHelp(),
+		Example: versionHelp(),
+		Run: func(cmd *cobra.Command, args []string) {
+			cmd.Printf("charts-syncer version: %s\n", Version)
+		},
```
jotadrilo, line 22: Please fix the indentation!

cmd/version.go
```diff
@@ -0,0 +1,26 @@
...
+}
+
+func newVersionCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "version",
+		Short:  versionHelp(),
+		Long: versionHelp(),
+		Example: versionHelp(),
+		Run: func(cmd *cobra.Command, args []string) {
+			cmd.Printf("charts-syncer version: %s\n", Version)
```
jotadrilo, line 21: I would just print the version

jotadrilo (approved):
LGTM! I added some nit-picks only

.goreleaser.yml
jotadrilo: nice!

cmd/version.go
```diff
@@ -0,0 +1,26 @@
+package cmd
+
+import (
+	"fmt"
+	"github.com/spf13/cobra"
```
Javirln, line 5: Add a new line between `fmt` and `github.com/spf13/cobra` to be consistent.

cmd/version.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+var Version = "dev"
```
jiparis, line 8: (nit) if this var being exported for any reason? 
If it's not, I would prefer it to be declared inside `newVersionCmd` to avoid collisions with other variables in the same package.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/bitnami-labs/charts-syncer
 
-go 1.15
+go 1.16
```
Javirln, line 3: Any reason no to use the 1.17 for example (the last one)? It's true it's not the scope of this PR, just thinking out loud

Javirln (commented):
LGTM!

migmartri (approved):

cmd/version.go
migmartri: @tompizmor if you don't have it you should probably add formatting on save in goland

migmartri:
Something unrelated

goreleaser is very cool, I didn't know about it. 

I've been checking the documentation though but couldn't find any information on `.goreleaser.yml`. How did you figure out (in the past not in this patch) what options were available for that tool?

cmd/version.go
jotadrilo: +1

jotadrilo:
check https://goreleaser.com/customization/build/

cmd/version.go
tompizmor: I thought I needed to export the variable in order to set the value form ldflags but it is not needed. I just changed it

cmd/version.go
tompizmor: Yeah! I just added it back. I would swear I already configured it a FileWatcher to run gofmt but it wasn't there 🤷‍♂️

go.mod
tompizmor: I tried that version but found a weird error in the tests only happening with go 1.17.
I would try to updated in another PR later

PR #117, opened 2021-12-09
chore: Extract OCI index library to its own package

jotadrilo:
In order to improve the readability of the OCI extension, let's move the logic related to the OCI index to its own package. 

### Changeset

- A new `useRemoteIndex` configuration field so the use of a remote index for OCI is declarative
- Rename `chartsIndex` configuration field to `remoteIndex` so it is aligned with the configuration above.
- The index filename is inferred now (instead of using a hardcoded name) from the layers matching known media types.

#### Enable remote index use

```yaml
source:
  repo:
    kind: OCI
    url: https://registry.pivotal.io/tac-for-tanzu-advanced/charts
    useRemoteIndex: true
target:
  repoName: myrepo
  containerRegistry: demo.goharbor.io
  containerRepository: jotadrilo/charts
  repo:
    kind: LOCAL
    path: /tmp/charts-syncer-test
```
```bash
❯ ./dist/charts-syncer sync --config config.yaml -v5
I1209 17:23:11.251520   28695 sync.go:32] Using config file: "config.yaml"
I1209 17:23:11.252496   28695 syncer.go:104] Using workdir: "/Users/jdrios/.charts-syncer"
I1209 17:23:11.252537   28695 cache.go:38] Allocating cache dir: "/Users/jdrios/.charts-syncer/134506542131cf0da32837b667f98697050c24a8"
I1209 17:23:11.252564   28695 oci.go:425] Attempting to retrieve remote index...
I1209 17:23:11.252586   28695 oci.go:108] Remote index was not provided, using default location: "registry.pivotal.io//tac-for-tanzu-advanced/charts/index:latest"
INFO[0001] trying next host - response was http.StatusNotFound  host=registry.pivotal.io
W1209 17:23:12.623212   28695 oci.go:442] The remote index does not exist yet. This process might be slow.
I1209 17:23:12.623239   28695 cache.go:38] Allocating cache dir: "/Users/jdrios/.charts-syncer"
W1209 17:23:12.623445   28695 sync.go:32] There were some problems loading the information of the requested charts: not found charts to sync
I1209 17:23:12.623469   28695 sync.go:47] There are no charts out of sync!
```

#### Enable remote index use with custom reference

```yaml
source:
  repo:
    kind: OCI
    url: https://registry.pivotal.io/tac-for-tanzu-advanced/charts
    useRemoteIndex: true
    remoteIndex: registry.pivotal.io/tac-for-tanzu-advanced/index:latest
target:
  repoName: myrepo
  containerRegistry: demo.goharbor.io
  containerRepository: jotadrilo/charts
  repo:
    kind: LOCAL
    path: /tmp/charts-syncer-test
```
```bash
❯ ./dist/charts-syncer sync --config config.yaml -v5
I1209 17:21:39.635418   28645 sync.go:32] Using config file: "config.yaml"
I1209 17:21:39.636141   28645 syncer.go:104] Using workdir: "/Users/jdrios/.charts-syncer"
I1209 17:21:39.636181   28645 cache.go:38] Allocating cache dir: "/Users/jdrios/.charts-syncer/134506542131cf0da32837b667f98697050c24a8"
I1209 17:21:39.636207   28645 oci.go:425] Attempting to retrieve remote index...
I1209 17:21:39.636228   28645 oci.go:110] Remote index was provided: "registry.pivotal.io/tac-for-tanzu-advanced/index:latest"
I1209 17:21:49.678542   28645 cache.go:38] Allocating cache dir: "/Users/jdrios/.charts-syncer"
I1209 17:21:49.678712   28645 index.go:85] Publishing threshold set to "1970-01-01 01:00:00 +0100 CET"
I1209 17:21:49.678724   28645 index.go:96] Found 1 versions for "airflow" chart: [11.1.9]
I1209 17:21:49.678736   28645 index.go:97] Indexing "airflow" charts...
...
```


tompizmor:
Why do you want to explicitly enable the index lookup? Can't we just keep the `remoteIndex` if we need to specify a non-default location?

jotadrilo:
> Why do you want to explicitly enable the index lookup? Can't we just keep the `remoteIndex` if we need to specify a non-default location?

I think the remote index is an artificial artifact (not officially supported by the community). Hence, I think it must be treated as experimental and, therefore, enabled purposely. WDYT?

tompizmor:
> > Why do you want to explicitly enable the index lookup? Can't we just keep the `remoteIndex` if we need to specify a non-default location?
> 
> I think the remote index is an artificial artifact (not officially supported by the community). Hence, I think it must be treated as experimental and, therefore, enabled purposely. WDYT?

Ok! Yes, I think if someone wants to use a chart index for OCI repositories he/she should know the expected index format and how to enable the feature. Otherwise, we could potentially try to use a random index artifact living in the repo but that is not compatible with charts-syncer. Hence, I agree on making it explicit.

internal/indexer/oci.go
```diff
@@ -0,0 +1,201 @@
...
+		return nil, errors.Wrapf(ErrInvalidArgument, "invalid OCI host URL: %+v", err)
+	}
+	resolver := newDockerResolver(u, opt.username, opt.password, opt.insecure)
+
+	ind := &ociIndexer{
+		reference: opt.reference,
+		resolver:  resolver,
+	}
+	if ind.reference == "" {
+		host := strings.Trim(strings.Join([]string{u.Host, u.Path}, "/"), "/")
```
tompizmor, line 106: This is not a proper host right? it also includes the project name and subpath? 
In that case I would avoid naming the variable as `host`, maybe `partialReference` or `partialPath`?

internal/indexer/oci.go
```diff
@@ -0,0 +1,201 @@
...
+// Get implements Indexer
+func (ind *ociIndexer) Get(ctx context.Context) (idx *api.Index, e error) {
+	// Allocate folder for temporary downloads
+	dir, err := os.MkdirTemp("", "indexer")
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to create temporary indexer directory")
+	}
+	defer func() {
+		err := os.RemoveAll(dir)
+		if e == nil {
```
tompizmor, line 132: why do you have this `e` error variable?

internal/indexer/oci.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	return idx, nil
+}
+
+func (ind *ociIndexer) downloadIndex(ctx context.Context, rootPath string) (f string, e error) {
+	// Pull index files from remote
+	store := content.NewFileStore(rootPath)
+	defer func() {
+		err := store.Close()
+		// This library is buggy, and we need to check the error string too
```
tompizmor, line 160: can you add more details about the error? Were we already using this library before?

internal/indexer/oci.go
jotadrilo: When a deferred function returns an error as part of the signature, we should always handle that case. This is pretty common with file closers (`defer f.Close`). A common pattern to handle the error of a deferred function so we can return handle the error outside is to use named return variables (`e error` in this case) so we can check the returned value (returned error) inside the deferred function body.

```
err := os.RemoveAll(dir)
		if e == nil {
```

Hence, here we are removing all directories (which might raise errors), then checking if the function has exited with errors (you have to decided whether to append or not the deferred function error here) and, if not, replace the returned error with your new error (or `nil` error, if everything went fine).

I learned this at https://www.joeshaw.org/dont-defer-close-on-writable-files/ so you may find it interesting too.

internal/indexer/oci.go
jotadrilo: this is the oras library, you can see how the `Close` function is producing empty errors:

```golang
# oras.land/oras-go/pkg/content/file.go:179

// Close frees up resources used by the file store
func (s *FileStore) Close() error {
	var errs []string
	s.tmpFiles.Range(func(name, _ interface{}) bool {
		if err := os.Remove(name.(string)); err != nil {
			errs = append(errs, err.Error())
		}
		return true
	})
	return errors.New(strings.Join(errs, "; "))
}
```

jotadrilo: I have filled an issue in the oras-go project: https://github.com/oras-project/oras-go/issues/84

tompizmor (approved):
LGTM

migmartri:
>     useRemoteIndex: true
Actually in this specific case we decided otherwise, to enable it by default so people that are already using and have configured chart-syncer get this functionality out of the box. Functionality that my understanding is that it does not affect the behavior of the tool (since the user will have filtering already set) but will get the performance gain.

I am ok with changing this decision though but personally if we are going to make somebody to indicate that they want to use an index I'd make them tell us what's the name of it.

README.md
```diff
@@ -151,32 +151,43 @@ target:
 It is worth mentioning that you can use Harbor robot accounts using OCI registries as source or target.
 
 Also, take into account that if you use OCI as the source repository you must specify the list of charts to synchronize
-or a pointer to an [index file](#helm-charts-index-for-oci-based-repositories) in the repository.
+or a pointer to a [remote index file](#remote-index-for-oci-based-repositories) in the repository.
 
-#### Helm Charts Index for OCI-based repositories
+#### Remote index for OCI-based repositories
```
migmartri, line 156: What's wrong with calling it Helm Charts index? I understand that's not an official solution and I believe that we make it clear somewhere else, but to me an index needs to indicate what's indexing,

README.md
```diff
@@ -151,32 +151,43 @@ target:
...
 charts-syncer will be able to auto-discover what charts need to be synchronized.
 
 ~~~yaml
 source:
  repo:
    kind: OCI
    url: https://my-oci-registry.io/my-project/subpath
-   chartsIndex: my-oci-registry.io/my-project/index:latest
+   # Enable remote index lookup
+   useRemoteIndex: true
```
migmartri, line 167: I still believe that charts index is a better name, it indicates what the index will contain and at the end of the day `remote` is implicit in the URI

internal/indexer/oci.go
```diff
@@ -0,0 +1,204 @@
...
+type ociIndexerOpts struct {
+	reference string
+	url      string
+	username string
+	password string
+	insecure bool
+}
+
+// OciIndexerOpt allows setting configuration options
+type OciIndexerOpt func(opts *ociIndexerOpts)
```
migmartri, line 32: nice!

internal/indexer/oci.go
```diff
@@ -0,0 +1,204 @@
...
+	} else {
+		klog.Infof("Remote index was provided: %q", ind.reference)
+	}
+
+	return ind, nil
+}
+
+// VacAssetIndexLayerMediaType is a media type used in VAC to store a JSON containing the index of
+// charts and containers in a repository
+const VacAssetIndexLayerMediaType = "application/vnd.vmware.tac.asset-index.layer.v1.json"
```
migmartri, line 118: do we really need to depend on this media type? I'd not like to have a VAC reference here if it's not strictly needed.

migmartri (commented):
This is nice!

I am not sure though if we want to rely on media types here, it will couple this code to VAC and will make it brittle since from that point onwards we will not need to make sure that the content meets a specific (documented proto) API but that it needs to be stored with some custom mimetypes encoded in the code itself.

This will prevent other people from adding their own index unless they identify it as a vac index right? Is that what we want?

Thanks!

jotadrilo:
> This is nice!
> 
> I am not sure though if we want to rely on media types here, it will couple this code to VAC and will make it brittle since from that point onwards we will not need to make sure that the content meets a specific (documented proto) API but that it needs to be stored with some custom mimetypes encoded in the code itself.
> 
> This will prevent other people from adding their own index unless they identify it as a vac index right? Is that what we want?
> 
> Thanks!

You hit the nail on the head. We are currently coupling this code to VAC in the most obscure way: relying on the filename that was uploaded by a workflow triggered by VAC itself.

The only way that you can make this generic is:
1. Document the media type expected by this tool:
  - Generic: `applications/json`
  - Specific: `application/vnd.vmware.charts-syncer.index.layer.v1.json`
2. Retrieve the original name of the index JSON file that a user uploaded from annotations (we are hardcoding `asset-index.json` because the VAC workflow is uploading the index from a filename called like that)

Hence, we are now in a better shape since the tool will allow generics.

internal/indexer/oci.go
jotadrilo: So far, it is the only way of controlling the layer that defines the index content and filename. Ideally, I would not do this either. See my other comment.

PR #119, opened 2021-12-10
feature: allow generating intermediate bundles for air gap scenarios

tompizmor:
In order to support air gap scenarios the first step is to generate the intermediate bundles containing the original chart, the container images and the container images hints file.

This PR adds a new repository kind called LOCAL_INTERMEDIATE_BUNDLE that can be used as target only for the moment (A future PR will implement the changes to use it a source too)

By using this repo as target, the output will be a tarball per chart in the defined repo path:

```
target:
  repo:
    kind: LOCAL_INTERMEDIATE_BUNDLE
    path: /Users/tpizarro/Desktop/workspace/charts-syncer-ws/bundles
```

**Output folder**
```
ls ~/Desktop/workspace/charts-syncer-ws/bundles

apache-8.9.3.tar
```

**Intermediate bundle content**
If we inspect that tarball, this is its expected content:

```
.
├── hints.yaml
├── images.tar
└── original-chart
```

**Example of use for apache sync**

```
$ ./dist/charts-syncer sync --config ./dist/config-chartmuseum-bundle-relok8s.yaml

I1210 14:33:04.245709   55319 sync.go:32] Using config file: "./dist/config-chartmuseum-bundle-relok8s.yaml"
I1210 14:33:04.276591   55319 sync.go:45] There are 2 charts out of sync!
I1210 14:33:04.276607   55319 sync.go:55] Syncing "common-1.10.1" chart...
.relok8s-images.yaml hints file found
Computing relocation...

Writing hints.yaml...
Writing Helm Chart files at original-chart/...
Writing 0 images...
E1210 14:33:04.278755   55319 sync.go:138] unable to move chart common:1.10.1: failed archiving images: failed to pack images: unable to calculate manifest: set of images is empty
I1210 14:33:04.278780   55319 sync.go:55] Syncing "apache-8.9.3" chart...
.relok8s-images.yaml hints file found
Computing relocation...

Writing hints.yaml...
Writing Helm Chart files at original-chart/...
Processing image index.docker.io/bitnami/apache:2.4.51-debian-10-r39
Processing image index.docker.io/bitnami/git:2.34.0-debian-10-r0
Processing image index.docker.io/bitnami/apache-exporter:0.10.1-debian-10-r56
Writing 3 images...
Intermediate bundle complete at /var/folders/23/7yxk2qds2rndbq5t5_r1wnzh0000gp/T/charts-syncer895343797/apache-8.9.3.intermediate.tar
Error: unable to move chart "common-1.10.1" with relok8s: failed archiving images: failed to pack images: unable to calculate manifest: set of images is empty
```
As you can see, it fails for the common chart as there are no images.


pkg/client/local_intermediate_bundle/local_intermediate_bundle.go
```diff
@@ -0,0 +1,132 @@
+package intermediate
```
tompizmor, line 1: this package is practically the same as the local one, just changing some filenames and extensions..

jotadrilo: why do we need a new configuration and a new package?

jotadrilo:
I do probably not understand the approach completely.

I think you are reusing the `sync` action to allow "syncing" a chart (and images, recursively) into a local directory. The main use of this kind of all-in-one bundle is air-gapped environments, which are supposed to be isolated from the public/private networks and, hence, unable to perform pull operations to retrieve dependencies.

If this is right, I think "sync" is a bit abused here. I do understand now that you have defined a new local repository so you can bound a different logic depending on the kind of repository. However, I think this is not completely right. I foresee a different action to craft or package the charts (and images) into a tarball that later you can operate (e.g. pushing it to an OCI repo). The plot twist here is that "sync" is thought to synchronize chart repositories and this new operation is thought to create a bundle of charts and deps. They are different and putting both actions together seems not the right decision.

Instead, I have a different proposal but please feel free to correct me if you guys already had conversations around this topic and it is the right direction, though. I would create a "download" operation so that you can specify to also download container images. That may converge within the current logic (since "sync" also download charts) but that's a different topic. My point here is that I see different logical steps that we could concatenate if we split this tool correctly. For example:

```
charts-syncer download -c oci-repo-config.yaml -f charts-bundle.tar.gz
charts-syncer upload -c air-gapped-env-config.yaml -f charts-bundle.tar.gz
```

migmartri (approved):
Great! 

I am assuming that we will be leaving the documentation of this feature to the end, correct?

migmartri:
This is abandoned right?

tompizmor:
Closing this PR as this functionality was finally added at #120

PR #120, opened 2021-12-14
feature: support air gap scenarios in charts-syncer

tompizmor:
Full air gap support in this PR!

## Notable changes

1. There have been some re-organization in the packages
We went from this:

```
pkg
├── client
│   ├── chartmuseum
│   ├── core
│   ├── harbor
│   ├── helmclassic
│   ├── local
│   ├── oci
│   └── types
├── syncer
└── testdata
    └── charts
```

to this:

```
pkg
├── client
│   ├── intermediate
│   ├── repo
│   │   ├── chartmuseum
│   │   ├── harbor
│   │   ├── helmclassic
│   │   ├── local
│   │   └── oci
│   └── types
├── syncer
└── testdata
    └── charts
```

2. There is a new package called intermediate with the client for this implementation.

3. Source and Target configuration sections now supports a single string to indicate the intermediate bundles path. It is a oneof so only a repo or a path can be defined.

## Example of use for air gap environments:

**Download bundles**

Use a config file like this:

```
source:
  repo:
    kind: CHARTMUSEUM
    url: http://localhost:8080
target:
  intermediateBundlesPath: /Users/tpizarro/Desktop/workspace/charts-syncer-ws/bundles
charts:
  - apache
```

Execute charts-syncer:

```
$ charts-syncer sync --config ./config-download-bundles.yaml
```

**Move bundles**
Move the bundles from the target path to the air gapped machine however you want.

**Upload bundles**

Use a similar config file:

```
source:
  intermediateBundlesPath: /Users/tpizarro/Desktop/workspace/charts-syncer-ws/bundles
target:
  repoName: myrepo
  containerRegistry: demo.goharbor.io
  containerRepository: tpizarro/containers
  repo:
    kind: CHARTMUSEUM
    url: http://localhost:9090
```

Execute charts-syncer:

```
$ charts-syncer sync --config ./config-upload-bundles.yaml
```

> :warning: DISCLAIMER
> **I plan to add the documentation in a separated PR**
> **I plan to add tests later in a separated PR**


pkg/client/client.go
```diff
@@ -0,0 +1,29 @@
+package client
+
+import (
+	"github.com/bitnami-labs/charts-syncer/pkg/client/types"
+	"helm.sh/helm/v3/pkg/chart"
+)
+
+// Reader defines the methods that a ReadOnly chart client should implement.
+type Reader interface {
```
josvazg, line 9: This is a ChartReader right? it does not read bytes or other things, it reads charts.

Maybe we should not change it in this PR, but I wonder if we should change it eventually.

pkg/client/client.go
```diff
@@ -0,0 +1,29 @@
...
+	Has(name string, version string) (bool, error)
+	GetChartDetails(name string, version string) (*types.ChartDetails, error)
+
+	// Reload reloads or refresh the client-side data, in case it needs it
+	Reload() error
+}
+
+// Writer defines the methods that a WriteOnly chart client should implement.
+type Writer interface {
+	Upload(filepath string, metadata *chart.Metadata) error
```
josvazg, line 22: Same comment for Writer, this is a `ChartsWriter` and probably `Upload` is more like a `Copy` in general.

Again, maybe for another PR.

pkg/client/client.go
```diff
@@ -0,0 +1,29 @@
...
+}
+
+// ReadWriter defines the methods that a chart client should implement.
+type ReadWriter interface {
+	Reader
+	Writer
+}
```
josvazg, line 29: Following the above reasoning this would become `ChartsReadWriter`.

pkg/client/intermediate/intermediate.go
```diff
@@ -0,0 +1,139 @@
...
+)
+
+var (
+	versionRe = regexp.MustCompile("(.*)-(\\d+\\.\\d+\\.\\d+)\\.bundle.tar")
+)
+
+// Repo allows to operate a chart repository.
+type Repo struct {
+	dir     string
+	entries map[string][]string
```
josvazg, line 28: Why is each map entry mapping to an array of strings? what is the key and the value of this map really?

josvazg: Oh I see, several versions of the same chart...

Maybe we should use a type to document that:
~~~go
type chartVersions []string
...
type Repo struct {
	dir     string
	entries map[string]chartVersions
...
~~~

pkg/client/intermediate/intermediate.go
```diff
@@ -0,0 +1,139 @@
...
+
+	input, err := ioutil.ReadFile(filepath)
+	if err != nil {
+		return errors.Annotatef(err, "reading %q", filepath)
+	}
+
+	out := path.Join(r.dir, fmt.Sprintf("%s-%s.bundle.tar", name, version))
+	if err := ioutil.WriteFile(out, input, 0644); err != nil {
+		return errors.Annotatef(err, "creating %q", out)
+	}
```
josvazg, line 120: How big is what is being read at `filepath`?
I would rather copy streams...
```suggestion
	src, err := os.Open(filepath)
	if err != nil {
		return errors.Annotatef(err, "reading %q", filepath)
	}

	out := path.Join(r.dir, fmt.Sprintf("%s-%s.bundle.tar", name, version))
	dst, err = os.Create(out, 0644)
	if err != nil {
		return errors.Annotatef(err, "creating %q", out)
	}
	_, err = io.Copy(dst, src)
	if err != nil {
		return errors.Annotatef(err, "copying %q to %q", filepath, out)
	}
```

pkg/client/intermediate/intermediate.go
```diff
@@ -0,0 +1,139 @@
...
+	sort.Strings(r.entries[name])
+
+	return nil
+}
+
+// GetChartDetails returns the details of a chart
+func (r *Repo) GetChartDetails(name string, version string) (*types.ChartDetails, error) {
+	return &types.ChartDetails{
+		PublishedAt: utils.UnixEpoch,
+		Digest:      "deadbeef",
```
josvazg, line 132: why is this always fake/empty?

pkg/client/intermediate/intermediate.go
```diff
@@ -0,0 +1,139 @@
...
+	}, nil
+}
+
+// Reload reloads the index
+func (r *Repo) Reload() error {
+	return nil
+}
```
josvazg, line 139: This looks like an interface mismatch. Seems this intermediate would better implement less methods.

pkg/client/intermediate/intermediate.go
josvazg: Is this another interface mismatch like Reload?

josvazg (commented):
Some questions...

api/config_test.go
```diff
@@ -9,22 +9,26 @@ import (
 func TestValidate(t *testing.T) {
 	config := &api.Config{
 		Source: &api.SourceRepo{
-			Repo: &api.Repo{
-				Url:  "ht//:fake.source.com",
-				Kind: api.Kind_CHARTMUSEUM,
-				Auth: &api.Auth{
-					Username: "user",
-					Password: "password",
+			Spec: &api.SourceRepo_Repo{
```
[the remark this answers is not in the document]
migmartri: Yeah, that's a good point, it might be worth renaming those keys since they are not directly marshalled

pkg/client/intermediate/intermediate.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/bitnami-labs/charts-syncer/pkg/client/types"
+	"helm.sh/helm/v3/pkg/chart"
+)
+
+var (
+	versionRe = regexp.MustCompile("(.*)-(\\d+\\.\\d+\\.\\d+)\\.bundle.tar")
+)
+
+// Repo allows to operate a chart repository.
+type Repo struct {
```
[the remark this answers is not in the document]
migmartri: I'd also indicate with a comment or an assignation the interface that they implement

```
var _ client.ReaderWriter = &Repo{}
```

pkg/client/intermediate/intermediate.go
```diff
@@ -0,0 +1,139 @@
...
+	dir     string
+	entries map[string][]string
+}
+
+// NewIntermediateClient returns a ReadWriter object
+var NewIntermediateClient = func(intermediateBundlesPath string) (client.ReadWriter, error) {
```
migmartri, line 32: out of curiosity why are you creating a variable and assigning an  Anonymous function instead of a regular named function?

pkg/client/intermediate/intermediate.go
```diff
@@ -0,0 +1,139 @@
...
+	return New(intermediateBundlesPath)
+}
+
+// New creates a Repo object from an api.Repo object.
+func New(dir string) (*Repo, error) {
```
migmartri, line 37: it's interesting that the entries get loaded in the constructor instead of the list method which to me makes more sense. In fact, the lest logic a constructor has the better.

To me, entries should be added when calling List similar to what a DB will do while getting queried

migmartri (commented):
Will finish the review tomorrow, looking good!

api/config.proto
```diff
@@ -13,12 +13,18 @@ message Config {
 
 // SourceRepo contains the required information of the source chart repository
 message SourceRepo {
```
migmartri, line 15: As @josvazg mentioned, it might make sense to rename this key name to Source so latter on the struct will be called Source_Repo

pkg/client/client.go
migmartri: It technically also reads relok8s bundles now but yes, that rename does not sound bad to me.

pkg/client/client.go
```diff
@@ -0,0 +1,29 @@
+package client
+
+import (
+	"github.com/bitnami-labs/charts-syncer/pkg/client/types"
+	"helm.sh/helm/v3/pkg/chart"
+)
+
+// Reader defines the methods that a ReadOnly chart client should implement.
```
migmartri, line 8: > chart client

This comment might be now outdated, it might make sense to add some more literature to this package documentation

pkg/client/repo/core.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/bitnami-labs/charts-syncer/internal/cache"
+	"github.com/bitnami-labs/charts-syncer/pkg/client/types"
+	"github.com/juju/errors"
+)
+
+// NewClient returns a Client object
+//
+// The func is exposed as a var to allow tests to temporarily replace its
+// implementation, e.g. to return a fake.
+var NewClient = func(repo *api.Repo, opts ...types.Option) (client.ReadWriter, error) {
```
migmartri, line 24: > // The func is exposed as a var to allow tests to temporarily replace its

this does not seem like a good idea, ideally mocking should be done via custom interfaces, in any case I understand that this was this way before so it's out of the scope of this patch to change it.

pkg/client/intermediate/intermediate.go
```diff
@@ -0,0 +1,139 @@
...
+func (r *Repo) List() ([]string, error) {
+	var names []string
+	for name := range r.entries {
+		names = append(names, name)
+	}
+	return names, nil
+}
+
+// ListChartVersions lists all versions of a chart
+func (r *Repo) ListChartVersions(name string) ([]string, error) {
```
migmartri, line 72: following the pattern I mentioned above, this method would call underneath list and then extract the versions.

pkg/client/intermediate/intermediate.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+	}
+	return false, nil
+}
+
+// Upload uploads a chart to the repo
+func (r *Repo) Upload(filepath string, metadata *chart.Metadata) error {
+	name := metadata.Name
+	version := metadata.Version
+	if _, ok := r.entries[name]; ok {
```
migmartri, line 104: what about getting this information from `ListChartVersions`? or even `Has`?

pkg/client/intermediate/intermediate.go
migmartri: If this method is never used it might make sense to return a NonImplemented error

pkg/syncer/sync.go
```diff
@@ -78,7 +78,8 @@ func (s *Syncer) SyncPendingCharts(names ...string) error {
 		}
 		var packagedChartPath string
 
-		if s.relocateContainerImages {
+		intermediateScenario := s.source.GetIntermediateBundlesPath() != "" || s.target.GetIntermediateBundlesPath() != ""
```
migmartri, line 81: I'd add a comment explaining that this means that we are running a partial sync, repo => intermediate or intermediate => repo

pkg/syncer/sync.go
```diff
@@ -112,33 +113,29 @@ func (s *Syncer) SyncPendingCharts(names ...string) error {
...
-					Path: outputChartPath,
-				},
-			},
-		},
+	packagedChartPath := ""
+	req := &mover.ChartMoveRequest{}
+	if s.target.GetIntermediateBundlesPath() != "" {
+		// First step of intermediate process
+		packagedChartPath = filepath.Join(outdir, fmt.Sprintf("%s-%s.bundle.tar", chart.Name, chart.Version))
+		req = s.getBundleRequest(chart.TgzPath, packagedChartPath)
```
migmartri, line 121: `getBundleRequest` `saveBundleRequest` might be more clear

pkg/syncer/sync.go
```diff
@@ -112,33 +113,29 @@ func (s *Syncer) SyncPendingCharts(names ...string) error {
...
-			Chart: mover.ChartSpec{
-				Local: &mover.LocalChart{
-					// output will be [chart-name]-[chart-version].relocated.tgz
-					Path: outputChartPath,
-				},
-			},
-		},
+	packagedChartPath := ""
+	req := &mover.ChartMoveRequest{}
+	if s.target.GetIntermediateBundlesPath() != "" {
```
migmartri, line 118: it might make sense to extract the logic that calculates the relok8s request to its own method so it's simpler to test.

```
func getRelok8sMoveRequest(source *Syncer.Source, target *Syncer.Target, chart *chart.Chart) *mover.ChartMoveRequest 
```

pkg/syncer/sync.go
```diff
@@ -200,3 +197,73 @@ func (s *Syncer) SyncWithChartsSyncer(ch *Chart, id, workdir, outdir string, has
...
+				Local: &mover.LocalChart{
+					Path: targetPath,
+				},
+			},
+		},
+	}
+	return req
+}
+
+func (s *Syncer) getBundleRequest(sourcePath, targetPath string) *mover.ChartMoveRequest {
```
migmartri, line 226: `get` makes it sounds like a generic getter, when in fact you want to get the request from the `get` action. There might be a better way to rename this, i.e

```
relok8sMoveReq
relok8sBundleSaveReq
relok8sBundleLoadReq
```

migmartri (commented):
This is cool! 

I'd take a look at how the new client implements the interface, loading the entries in the constructor does not look natural to me, specially if there is no way to "reload" it's content.

api/config.proto
jotadrilo: +1

internal/config/config.go
```diff
@@ -56,10 +56,10 @@ func Load(config *api.Config) error {
 			return err
 		}
 	}
-	if err := config.Source.Repo.SetBasicAuth(os.Getenv("SOURCE_AUTH_USERNAME"), os.Getenv("SOURCE_AUTH_PASSWORD")); err != nil {
+	if err := config.GetSource().GetRepo().SetBasicAuth(os.Getenv("SOURCE_AUTH_USERNAME"), os.Getenv("SOURCE_AUTH_PASSWORD")); err != nil {
```
jotadrilo, line 59: you are not checking if the source has a `repo` in this method

pkg/client/client.go
jotadrilo: +1

pkg/client/intermediate/intermediate.go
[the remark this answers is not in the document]
jotadrilo: +1 to both comments

pkg/client/intermediate/intermediate.go
jotadrilo: I think this allows overwriting in testing

pkg/client/intermediate/intermediate.go
jotadrilo: +1

pkg/client/intermediate/intermediate.go
jotadrilo: +1

pkg/client/intermediate/intermediate.go
```diff
@@ -0,0 +1,139 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
+	"regexp"
+	"sort"
+
+	"github.com/bitnami-labs/charts-syncer/pkg/client"
```
jotadrilo, line 12: reformat imports!

pkg/client/intermediate/intermediate.go
jotadrilo, line 26: Ideally, I would treat the intermediate "store" as a new object in the configuration:

```yaml
source:
  intermediate:
    kind: BUNDLES
    path: /path/to/bundles/dir
```

This way we should have `pkg/client/intermediate/intermediate.go` including a generic `New` constructor and then a `pkg/client/intermediate/bundles` with this logic. However, I am totally fine having this simplification today.

migmartri:
I tested out this PR and something I noticed is that in the first leg (repo => intermediateBundleDir) we also store the bundles for the subcharts. 

For example if I run the sync with the following configuration, I get the `memcached` bundles too.

```
source:
  repo:
    kind: HELM
    url: https://charts.app-catalog.vmware.com/demo
target:
  intermediateBundlesPath: "bundles"
charts:
  - "wordpress"
```

This is unnecessary since relok8s does not care about non vendored charts (yet). In fact when you run the second leg of the relocation (intermediate-bundle => Helm repo) those get ignored since they are filtered out (I initially just asked for wordpress charts)

pkg/client/client.go
tompizmor: Added something, but not really sure if that's enough

internal/config/config.go
tompizmor: right! however it is not failing 🤔 maybe the getter is managing the case of repo being nil?

pkg/client/intermediate/intermediate.go
tompizmor: I would skip it for now

pkg/client/intermediate/intermediate.go
tompizmor: There are other methods that relies on the `entries` info initialised in the constructor (ListChartVersions(), Has(), Upload())...moving the entries initialisation to the List() method means that we cannot call to the other methods before List() because entries would be empty or we will need to retrieve the info always from the List() method.

The later means that we would be reading the content of a folder several times per chart while having the entries populated just once seems more efficient to me. I would leave it as it is

tompizmor:
@josvazg, @jotadrilo, @migmartri  if you agree I don't want to add more things to this PR as it is already complex enough. However, once this PR gets landed I will work on the following PRs:

- Add tests
- Add documentation
- Rename interface names to something more meaningful
- Remove Reload() method from current interface as it is not used at all
- Allow skip dependency resolution when using relok8s. Due to the unexpected behaviour noticed by Migue [here](https://github.com/bitnami-labs/charts-syncer/pull/120#issuecomment-994874083)

jotadrilo:
I am OK!

go.mod
```diff
@@ -22,7 +22,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/spf13/cobra v1.2.1
 	github.com/spf13/viper v1.8.1
-	github.com/vmware-tanzu/asset-relocation-tool-for-kubernetes v0.3.45
+	github.com/vmware-tanzu/asset-relocation-tool-for-kubernetes v0.3.49
```
migmartri, line 25: Could you actually update to `v0.3.58` that one contains the fix for the duplicated assets

tompizmor: yep! just did it

josvazg:
> @josvazg, @jotadrilo, @migmartri if you agree I don't want to add more things to this PR as it is already complex enough. However, once this PR gets landed I will work on the following PRs:
> 
> * Add tests
> * Add documentation
> * Rename interface names to something more meaningful
> * Remove Reload() method from current interface as it is not used at all
> * Allow skip dependency resolution when using relok8s. Due to the unexpected behaviour noticed by Migue [here](https://github.com/bitnami-labs/charts-syncer/pull/120#issuecomment-994874083)

LGTM

migmartri (approved):
LGTM!

migmartri:
This is awesome BTW, great job!

PR #122, opened 2021-12-16
refactor: rename reader and writer interfaces

tompizmor:
Follow up from #120. We can use a more meaningful name for these interfaces


jotadrilo (approved):
LGTM!

migmartri (approved):
LGTM!

PR #123, opened 2021-12-16
test: add tests for new functions related to airgap support

tompizmor:
Added tests for new intermediate client added at #120 as well as some syncer methods to get the proper relok8s request.


migmartri (approved):
LGTM!

PR #124, opened 2021-12-17
feat: allow skip syncing dependencies

tompizmor:
This PR adds a new CLI flag --skip-dependencies to allow skip syncing the dependencies


migmartri (approved):
Maybe this option should be overridden in the airgap mode since as of today, this option makes no sense in that scenario and in fact it should be set to true by default (in contrast to the other modes)

Maybe if we detect that syncSkipDependencies is true and we are doing an airgap sync (or regular relok8s mode), we override it and show a warning message saying that the option has been ignored, that's not currently supported in intermediate targets.

tompizmor:
I was already overriding the option for intermediate scenarios but wasn't showing any warning. Now it is show and the direct relok8s sync also disables the dependency sync

migmartri (approved):
LGTM!

PR #127, opened 2022-01-03
fix: remove .relocated suffix when relok8s is the backend used

### LINKED ISSUE #125: Wrong resulting chart name
if the backend used is `relok8s` the resulting Helm Chart stored in the target repository is called *.relocated.tgz

```
$ oras pull     us-west1-docker.pkg.dev/bitnamigcetest2/relok8s-test/charts/wavefront-adapter-for-istio:1.0.10 -a
$ ls wavefront-adapter-for-istio*
wavefront-adapter-for-istio-1.0.10.relocated.tgz
```

We should probably drop the `.relocated` suffix

tompizmor:
Delete the .relocated suffix from chart name so in the target repository is stored as NAME-VERSION.tgz

Fix #125


migmartri (approved):
Thanks!

PR #131, opened 2022-01-14
feat: update charts index format

tompizmor:
Update charts-index format and default name of the index


tompizmor:
Not sure if we should make `useChartsIndex` config property true by default so no special configuration is needed for VAC repositories

jotadrilo (approved):
LGTM!

However, I think this PR deserves a better description and background. Please state that we would like to use a charts-specific index that contains all chart versions released internally, unequivocally.

migmartri (approved):
LGTM!

migmartri:
> Not sure if we should make useChartsIndex config property true by default so no special configuration is needed for VAC repositories

Yeah, same thing about relocateContainerImages. if we want to make this tool to work ootbox with TAC repositories those two option defaults should be set.

Maybe we can do it in another PR and document the reasoning wdyt?

migmartri (approved):

PR #135, opened 2022-01-19
feat: OCi charts-index retrieval by default

### LINKED ISSUE #132: Enable index and container sync options by default
Currently, we have two optional properties that are disabled by default that we could enable to get the best out of the box experience.

> useChartsIndex

This one could be enabled by default as long as we keep the current behavior which is to enhance the UX not making it a requirement in any case by falling back to the `charts:` filter.

> relocateContainerImages

The reason we left this option to false was so we do not change previous behavior of the tool. But there are other methods to communicate behavioral changes such as cutting a new release and so on.  

WDYT? 

cc/ @tompizmor @jotadrilo

migmartri:
Deprecates previous `useChartsIndex` replacing it an opt out alternative `disableChartsIndex`

In practice this makes the  OCI index enabled by default.

NOTE: This PR initially started covering the enablement of container images relocation by default too, but I decided to not to roll out that feature yet until we are ready documentation/UX wise.

Refs #132  

Signed-off-by: Miguel Martinez Trivino <mtrivino@vmware.com>


README.md
```diff
@@ -213,46 +217,35 @@ or a pointer to a [charts index file](#charts-index-for-oci-based-repositories)
...
-~~~
-
 By default, the library will look up for a "charts-index:latest" charts index artifact within the source OCI repository.
 
 However, this can be customized using the `chartsIndex` field using the format `REGISTRY/PROJECT/[SUBPATH][:TAG|@sha256:DIGEST]`.
 
+For example, if your URL is `https://my-oci-registry.io/my-project/subpath` and no `chartsIndex` is specified, charts-syncer will try to use 
+`my-oci-registry.io/my-project/subpath/charts-index:latest` asset as index if it exists.
+
+An example of the valid index format can be seen directly in its [protobuffer definition](internal/indexer/api/index.proto). Worth to mention
```
jotadrilo, line 227: Please use either `Protobuf` or `Protocol Buffer` from the docs:

```suggestion
An example of the valid index format can be seen directly in its [Protobuf definition](internal/indexer/api/index.proto). Worth to mention 
```

README.md
```diff
@@ -213,46 +217,35 @@ or a pointer to a [charts index file](#charts-index-for-oci-based-repositories)
...
+
 ~~~yaml
 source:
  repo:
    kind: OCI
    url: https://my-oci-registry.io/my-project/subpath
-   # Enable charts index lookup
-   useChartsIndex: true
-   chartsIndex: my-oci-registry.io/my-project/index:prod
+   # disableChartsIndex: true #
```
jotadrilo, line 235:
```suggestion
   # disableChartsIndex: true
```

README.md
```diff
@@ -213,46 +217,35 @@ or a pointer to a [charts index file](#charts-index-for-oci-based-repositories)
...
 ~~~yaml
-#
-# Example config file
-#
 source:
   repo:
     kind: OCI
     url: https://my-oci-registry.io/my-project/subpath
 ...
+# Required If not index is provided or found
```
jotadrilo, line 248:
```suggestion
# Required if no index is provided or found
```

jotadrilo (approved):
LGTM!

tompizmor (approved):
LGTM!

PR #140, opened 2022-01-26
feat: Add skipCharts property

migmartri:
Adds `skipCharts` property as an alternative to the opt-in `Charts` property.

This is specially useful for cases where you know that your Helm Chart is not going to successfully relocate, either because the container images are not accessible or even because it's not relok8s relocatable yet (missing relok8s metadata file)

This option still allows the user to make explicit decisions on what to skip while letting the current strict behavior of the tool and index based chart discovery to take place. 

Signed-off-by: Miguel Martinez Trivino <mtrivino@vmware.com>


tompizmor (approved):
LGTM! Thanks for implementing this so quickly!

PR #149, opened 2022-03-18
chore: update oras module

tompizmor:
- Update Oras module from v0.4.0 to v.1.1.0
- Update project to use go 1.17 (requirement of the oras 1.1.0 through [strconv.QuotedPrefix()](https://pkg.go.dev/strconv#QuotedPrefix) function usage)


internal/indexer/oci.go
```diff
@@ -158,33 +159,34 @@ func (ind *ociIndexer) downloadIndex(ctx context.Context, rootPath string) (f st
...
+		}),
 	}
-	_, layers, err := oras.Pull(ctx, ind.resolver, ind.reference, store, opts...)
+	// Fallback to the default index filename if the layers don't specify it
+	if indexFilename == "" {
+		klog.Infof("Unable to find index filename: using default")
+		indexFilename = defaultIndexFilename
+	}
+
+	_, err := oras.Copy(ctx, ind.resolver, ind.reference, store, ind.reference, opts...)
```
Javirln, line 183: Isn't this going to provoke a race condition on `indexFilename`?

tompizmor: yep, you are right! This code blocks order is not correct. We should do the oras.Copy() before checking if indexFilename is empty

Javirln (approved):
LGTM!

Issue #104, opened 2021-09-15
Make repository/registry replacement optional


maxvt-tonal:
Hi and thanks for offering the tool to the public! 

I am trying to fit it to a specific use case, and I'm concerned about the repository/registry replacement feature: to me it seems to assume both a very specific use case and a bitnami-specific convention of creating the Kubernetes image parameter from registry, repository, and image parts. Would you consider making repository/registry rewriting optional?

For the convention, the split between repository/registry/tag is not universally used and is not a Kubernetes primitive in the API. I've noticed this convention in many (if not all) bitnami charts, but it's nowhere to be seen in many other charts, and we don't use this convention currently. 

For charts that don't use this convention, I wouldn't want automatic rewriting to happen--but it is mandatory to specify it and if we use these keys they will get replaced--regardless of whether the chart follows this convention or not.

For the use case, it seems assumed by the design of this tool that a sync operation would only work on charts having the same repository (since only one replacement string is possible), probably on only a few versions of the same chart at a time. This is not, in the general case, the situation -- a museum can contain many local and 3rd-party charts, referencing many different images and repositories. Syncing the entire museum in this case is unlikely to produce a functioning copy.

Perhaps this is tailored to fit a specific bitnami's workflow, but this tool would be more generally applicable (for example, if I want to make an exact backup; if I want to reuse a registry across two museums, or if different registries are used at the same time; if I want to process a batch of charts using different repositories) if this replacement functionality was made optional. Would you consider such a change?

slushysnowman:
I'm having this issue at the moment, I'm looking to sync a ton of helm charts to our Harbor registry, and modify the images to be pulled from a proxy cache repo in Harbor. But the images all use different repos.

For example if you look at the [values for this chart](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/values.yaml)

Some images use `k8s.gcr.io/provider-aws/` as repo, some use `k8s.gcr.io/sig-storage/`

In my .yaml for the sync I have:
```
target:
  containerRegistry: $REGISTRY
  containerRepository: k8s-nspaas/provider-aws
  repo:
    kind: OCI
    url: $REGISTRY
```

But this means all the image repos get modified to use provider-aws, which results in the helm chart failing to deploy as it's the wrong repo for some of the images.

Is there a way around this?

marthydavid:
+1 for that.

[...]

PR #150, opened 2022-04-04
Add flag to skip rewriting values.yaml

LP0101:
Added a flag to skip rewriting the `values.yaml`

Usage:

`--skip-values`: Prevent charts-syncer from rewriting the image paths in values.yaml

Solves #104


jotadrilo:
Hello @LP0101, thanks for your contribution.

However, we would like to handle this issue using #153. The main reason is that, while this feature enables skipping the values rewrite via CLI, #153 enables skipping the container registry target definition. This is scoped and enough to fix #104 while it gives room to modify any, possibly, setting in the values YAML file unrelated to the container registry.

Sounds good?

PR #159, opened 2022-05-09
chore: update testing components version and charts-syncer flags

tompizmor:
Update old version of Helm, Kind, and Kubectl used for testing.
Additionally, just sync the latest version of the testing app (ghost) instead of syncing all available versions from specificdate.


jotadrilo (approved):
LGTM!

Issue #167, opened 2022-08-01
Chart indexing fails: Truncated gzip input


jotadrilo:
### Context

Today, I tried syncing a whole Helm Chart repository to a local directory. However, during the re-package process (uncompress source, indexed, chart, modify it, re-package it, and "upload" it to local) the sync failed with:

```
I0801 11:27:42.723174   72993 sync.go:171] Packaging "schema-registry-4.0.0"
I0801 11:27:42.756879   72993 sync.go:105] Uploading "schema-registry-4.0.0" chart...
Error: uncompressing "/Users/jdrios/.charts-syncer/4f2301d176a9788434844a4ad9839aac81ab7c84/airflow-11.1.6.tgz": unexpected EOF
```

If I try to uncompress the indexed chart manually I get the following error:

```
❯ tar xzf airflow-11.1.6.tgz
airflow/charts/postgresql/values.yaml: truncated gzip input
tar: Error exit delayed from previous errors.
```

### Steps to reproduce

This is the configuration file I used (I changed the source repo url here):

```yaml
source:
  repo:
    kind: HELM
    url: https://acme.org/mychartrepo

target:
  containerRegistry: docker.io
  containerRepository: foo
  repo:
    kind: LOCAL
    path: ./local-repo
```

```
❯ ./charts-syncer version
0.18.0

❯ ./charts-syncer sync --config charts-syncer.yaml -v 4
```

jotadrilo:
I could not detect the conditions under which this process fails to generate the corrupted tarball.
However, I have spotted that we are not closing file descriptors (nor HTTP body responses) correctly so I have sent #168

PR #168, opened 2022-08-01
[GAL-2124] Invalidate cache on fetch errors

jotadrilo:
Signed-off-by: Joseda Rios <jdrios@vmware.com>

We detected problems during the charts indexing process (see #167).

However, while I could not detect the conditions under which this process fails to generate the corrupted tarball, I spotted that we are not closing file descriptors (nor HTTP body responses) correctly.

This patch refactors the OCI and Helm repos to reuse the same logic to fetch and cache charts using the HTTP protocol, and to correctly close file descriptors and responses. It also removes the cached chart in case of an error writing/closing the file. This will avoid unexpected issues during the next sync.

Additionally, it moves the disk cache implementation to a subpackage to decouple the interface from the implementation.

New runs of the sync process haven't generated corrupted tarballs, but I cannot ensure this patch is the right bug fix.


internal/cache/cachedisk/cachedisk.go
```diff
@@ -0,0 +1,122 @@
...
+	klog.V(4).Infof("cache hit { op:has, id:%s, filename:%s }", c.id, filename)
+	return ok
+}
+
+// Read reads a cached file named by filename and writes the contents to the
+// provider w writer.
+//
+// If the file does not exist, Read returns an error that satisfies
+// errors.IsNotFound().
+func (c *Cache) Read(w io.Writer, filename string) error {
```
tompizmor, line 64: I don't see where you are using this method...

internal/cache/cachedisk/cachedisk.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	klog.V(4).Infof("cache hit { op:read, id:%s, filename:%s }", c.id, filename)
+	return nil
+}
+
+// Store stores the provided bytes in a cache file named by filename
+//
+// If the file exists, Store returns an error that satisfies
+// errors.IsAlreadyExists(). Otherwise, Store creates it with permissions perm
+func (c *Cache) Store(r io.Reader, filename string) error {
```
tompizmor, line 83: same...is this being used?

internal/cache/cachedisk/cachedisk.go
jotadrilo: I am not using this in the code that I have modified. Note that this file was moved from `cache/store.go`

internal/cache/cachedisk/cachedisk.go
jotadrilo: same

PR #175, opened 2022-12-14
Rewrite implementation of Has() method in oci.Repo

robinkb:
For our use case, we want to sync Helm charts between Sonatype Nexus, a classic Helm repo, to Harbor, an OCI registry. This took over three hours to complete, which lead me to investigate what it was doing that was taking so long.

For every version of a chart that it finds on Nexus, it checks if that version already exists on Harbor. To do that, it lists all of that chart’s versions by downloading every single version that exists in Harbor for that chart. For one of our charts, we have 611 versions in Nexus. charts-syncer was doing 611 * 611 = 373.321 GET calls to Harbor. It spent three hours just indexing the charts.

I rewrote the implementation of the Has() method in oci.Repo so that it instead performs a single HEAD request to check if the version is present on the registry. That seems to work just fine for us.

The tests fail because the test server doesn't handle HEAD requests, and I couldn't figure out how to fix that. Changing it to a GET request fixes the test, but that's not the most optimal choice performance-wise.

### DIFF
--- a/pkg/client/repo/oci/oci.go
+++ b/pkg/client/repo/oci/oci.go
@@ -272,7 +272,9 @@ func (r *Repo) Has(name string, version string) (bool, error) {
 	defer cancel()
 
 	u := *r.url
-	// Form API endpoint URL from repo url
+	// Form API endpoint URL from repo url as per the specification:
+	// https://github.com/opencontainers/distribution-spec/blob/main/spec.md#checking-if-content-exists-in-the-registry
+	// A successful request should return 200 OK.
 	u.Path = path.Join("v2", u.Path, name, "manifests", version)
 	req, err := http.NewRequestWithContext(ctx, "HEAD", u.String(), nil)
 	if err != nil {
--- a/pkg/client/repo/oci/oci.go
+++ b/pkg/client/repo/oci/oci.go
@@ -274,7 +274,7 @@ func (r *Repo) Has(name string, version string) (bool, error) {
 	u := *r.url
 	// Form API endpoint URL from repo url as per the specification:
 	// https://github.com/opencontainers/distribution-spec/blob/main/spec.md#checking-if-content-exists-in-the-registry
-	// A successful request should return 200 OK.
+	// The request should return 200 OK if the manifest exists.
 	u.Path = path.Join("v2", u.Path, name, "manifests", version)
 	req, err := http.NewRequestWithContext(ctx, "HEAD", u.String(), nil)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


tompizmor:
Hi @robinkb! Thanks for debugging the problem and highlighting the issue. As you pointed out the current implementation is suboptimal since the number of requests slows down the process. I think the proposed solution is fine. Let me trigger the CI and run some other tests.

pkg/client/repo/oci/oci.go
```diff
@@ -268,14 +268,35 @@ func (r *Repo) Fetch(name string, version string) (string, error) {
 
 // Has checks if a repo has a specific chart
 func (r *Repo) Has(name string, version string) (bool, error) {
-	versions, err := r.ListChartVersions(name)
+	ctx, cancel := context.WithTimeout(context.Background(), getTimeout)
+	defer cancel()
+
+	u := *r.url
+	// Form API endpoint URL from repo url
```
jotadrilo, line 275: can you please add a reference to the specification? Specifically to the [_Checking if content exists in the registry_](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#checking-if-content-exists-in-the-registry) section

jotadrilo:
> Hi @robinkb! Thanks for debugging the problem and highlighting the issue. As you pointed out the current implementation is suboptimal since the number of requests slows down the process. I think the proposed solution is fine. Let me trigger the CI and run some other tests.

we should probably help @robinkb to enhance the test server to handle HEAD requests.

pkg/client/repo/oci/oci.go
jotadrilo: curiously, we added the initial implementation at the same time that the docs around this were merged (2y ago).

I remember we actually discussed the sub-optimal performance of this method due to not having a way of just checking for specific artifacts. It is nice that you took the time to look into it and fix it! 😄

robinkb: Should be done!

And you're welcome! We are using this application to sync our Helm charts, and it works great. Only this one method needed some work :)

robinkb:
@jotadrilo I took a little bit more time and modified the test server as well. Is that solution okay?

tompizmor:
@robinkb I think it would be more correct if we create a different block just for the HEAD request since in this case, we should not return any manifest. Hence I would add something like:

```
	if ociTagManifestRegex.Match([]byte(r.URL.Path)) && r.Method == "HEAD" {
		testBasicAuth(rt.t, r)
		rt.HeadManifest200(w)
		return
	}
```

And I would also create the HeadManifest200() function next to the HeadManifest404() one.

robinkb:
Sorry for the delay! Completely fell off my radar. Please take another look.

tompizmor:
Hi @robinkb! Thanks for taking the time to update the PR! I will merge it and release a new version.

PR #181, opened 2023-05-03
Properly discover existing image in OCI target

makkes:
When using ghcr.io as an OCI target then charts-syncer would always overwrite an already existing chart because ghcr.io requires a client to retrieve a bearer token before making any requests against it. This is also true when basic auth credentials are provided. The `Repo#Has` method would always fail in that case because it can't handle 401 responses.

I exchanged the code in the `Has` method to use the go-containerregistry client instead that properly handles authentication.

# Before

```
$ go run main.go sync --latest-version-only
I0503 18:44:41.867762 3171500 sync.go:49] Looking for the default config charts-syncer.yaml
I0503 18:44:42.448242 3171500 sync.go:47] There is 1 chart out of sync!
I0503 18:44:42.448269 3171500 sync.go:55] Syncing "grafana-6.56.1" chart...
```

# After

```
$ go run main.go sync --latest-version-only
I0503 18:44:16.945449 3168464 sync.go:49] Looking for the default config charts-syncer.yaml
I0503 18:44:17.933216 3168464 sync.go:49] There are no charts out of sync!
```

### DIFF
--- a/pkg/client/repo/oci/oci.go
+++ b/pkg/client/repo/oci/oci.go
@@ -18,6 +18,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
 	"github.com/juju/errors"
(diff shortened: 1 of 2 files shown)


makkes:
@jotadrilo @tompizmor any chance to get this in? 🙏🏻

makkes:
Thanks @tompizmor, rebased and pushed.

PR #196, opened 2023-11-03
fix: replace deprecated iotil functions with os/io equivalents

wilmardo:
When I was creating the other PR my IDE complained about the ioutil pkg being deprecated.
> Deprecated: As of Go 1.16, the same functionality is now provided by package [io](https://pkg.go.dev/io) or package [os](https://pkg.go.dev/os), and those implementations should be preferred in new code. See the specific function documentation for details.

Looked at the [docs](https://pkg.go.dev/os) and all them simply call an other function from the `os` or `io` pkg. Replicated the deprecated functions with their newer equivalent.

### DIFF
--- a/internal/cache/cachedisk/cachedisk.go
+++ b/internal/cache/cachedisk/cachedisk.go
@@ -2,7 +2,6 @@ package cachedisk
 
 import (
 	"io"
-	"io/ioutil"
 	"os"
 	"path"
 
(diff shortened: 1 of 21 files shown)


jotadrilo (approved):
LGTM!

A perfect example of a camping rule!

tompizmor (approved):
Thanks for the MR!! 🙂

PR #231, opened 2025-02-07
github-actions: Add gchat notification

migruiz4:
### Description of the change

This PR adds GChat notifications to the CD pipeline.

When the pipeline fails, it will notify the team with a message using the gchat-notification worflow from bitnami/support

### Possible drawbacks

None known.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,18 +1,19 @@
 name: CI
 on:
   workflow_dispatch:
-  push: 
+  push:
     branches:
        - v2
   pull_request:
     branches:
       - v2
     types:
-      - assigned 
-      - opened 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -1,24 +1,24 @@
...
-
+# Remove all permissions by default
+permissions: {}
 env:
   KIND_VERSION: "0.12.0"
   KUBECTL_VERSION: "1.22.0"
-
 jobs:
-  Validate:
+  validate:
```
fmulero: Just a warn. Protection rules should be changed also

fmulero (dismissed):
LGTM. Be careful with protection rules, those will be affected by the change in the job name

PR #256, opened 2025-08-18
feat: Bump Go version to 1.25

juan131:
### Description of the change

Bump Go version to `1.25`.

### Benefits

Keep the project updated.

### Possible drawbacks

None


internal/utils/utils.go
```diff
@@ -64,8 +64,8 @@ func ChartExistInIndex(name string, version string, index *helmRepo.IndexFile) b
 	if index.Entries[name] != nil {
 		klog.V(3).Infof("Chart %q exists in index.yaml file. Searching %q version", name, version)
 		for i := range index.Entries[name] {
-			if index.Entries[name][i].Metadata.Version == version {
-				klog.V(3).Infof("Version %q found for chart %q in index.yaml file", index.Entries[name][i].Metadata.Version, name)
+			if index.Entries[name][i].Version == version {
```
tompizmor, line 67: Is this change related to the go version bump?

pkg/client/target/local/local.go
```diff
@@ -25,5 +25,5 @@ func New(dir string) (*Target, error) {
 // Unwrap unwraps a chart. In local mode, we do not actually unwrap, we just copy over the file as
 // we do not have a registry to write into the images and relocate
 func (t *Target) Unwrap(file string, metadata *chart.Metadata, _ ...config.Option) error {
-	return t.Repo.Upload(file, metadata)
+	return t.Upload(file, metadata)
```
tompizmor, line 28: same question

internal/utils/utils.go
juan131: It's because of the linter:
```
internal/utils/utils.go:67:30: QF1008: could remove embedded field "Metadata" from selector (staticcheck)
			if index.Entries[name][i].Metadata.Version == version {
```

pkg/client/target/local/local.go
juan131: Same here:
```
pkg/client/target/local/local.go:28:11: QF1008: could remove embedded field "Repo" from selector (staticcheck)
	return t.Repo.Upload(file, metadata)
```

tompizmor (approved):

internal/utils/utils.go
tompizmor: Perfect! Didn't know that was supported

PR #280, opened 2025-10-10
feature: use buf to generate code

tompizmor:
### Description of the change

This PR replace the deprecated `prototool` tool to generate code in favor of its modern `buf` replacement.

### Benefits

Stop using a deprecated tool for the development workflow.

### Possible drawbacks

None

### Applicable issues

None


internal/config/config.go
```diff
@@ -24,7 +24,49 @@ const DefaultIndexName = "charts-index"
 // DefaultIndexTag is the tag for the OCI artifact with the index
 const DefaultIndexTag = "latest"
 
-func setDefaultChartsIndex(config *api.Config) error {
+// Validate validates the config file is correct
+func Validate(c *apiv1.Config) error {
+	if repo := c.GetSource().GetRepo(); repo != nil {
+		switch k := repo.GetKind(); k {
+		case apiv1.Kind_CHARTMUSEUM, apiv1.Kind_HELM, apiv1.Kind_HARBOR, apiv1.Kind_OCI:
```
fjagugar: Is there any kind where the url doesn't need to be a valid url? Why not performing this validation always?

internal/config/config.go
```diff
@@ -24,7 +24,49 @@ const DefaultIndexName = "charts-index"
...
+		// so the user does not need to set it up
+		if auth.Username == "" || auth.Password == "" {
+			return errors.Errorf(`"target.containers.auth" "username"" and "password" are required"`)
+		}
+	}
+	if repo := c.GetTarget().GetRepo(); repo != nil {
+		if repo.GetKind() != apiv1.Kind_OCI && repo.GetKind() != apiv1.Kind_LOCAL {
+			return errors.Errorf(`"target.repo.kind" should be "OCI" or "LOCAL"`)
+		}
+	}
```
fjagugar: Why not unifying this validation with the validation we have above about the target repo?

internal/config/config.go
```diff
@@ -24,7 +24,49 @@ const DefaultIndexName = "charts-index"
...
+
+	return nil
```
fjagugar: Should we append all errors and return a single error with all the messages?

internal/config/config.go
tompizmor: Yes, LOCAL kind does not need a url but a path

fjagugar: Then maybe the logic should be "if not LOCAL, then it should be a valid URL", just in case in the future we add some extra kind different from these ones and we forget to add it to the validation.

internal/config/config_test.go
```diff
@@ -5,14 +5,48 @@ import (
...
+					Password: "password",
+				},
+			},
+		},
+	}
+
+	if err := Validate(config); err == nil {
+		t.Errorf("expected error but got nothing")
+	} else {
+		expectedError := `"source.repo.url" should be a valid URL: parse "ht//:fake.source.com": invalid URI for request`
```
fjagugar: Should we add more than one error to validate that they get merged?

fjagugar (dismissed):

internal/config/config_test.go
tompizmor: done

fjagugar (approved):

PR #282, opened 2025-10-10
Clean interfaces and remove unused methods

tompizmor:
### Description of the change

This PR removes some unused methods from the main interfaces. In particular, the Upload() method is unused since we started using the helm-distribution-for-helm library


fjagugar (commented):
Please check the linter errors.

fjagugar (approved):

